PR #6, opened 2018-03-25
feat: Support Angular propDecorators inputs and outputs

getsaf:
There is yet *another* way to add inputs on an Angular directive. I found that `FormConrolDirective` used a `propDecorators` object. Looking at the [source](https://github.com/angular/angular/blob/master/packages/forms/src/directives/reactive_directives/form_control_directive.ts#L144), I don't see anything special about what they are doing but for whatever reason, the inputs/outputs show up under the `<DirectiveClass>.propDecorators` object.

~~TODO: Should probably consider automatically hooking up output emitters on directives like it's done with components. If people are using `@angular/forms` junk, they'll likely run into it for things like `(ngOnChange)`.~~

### DIFF
--- a/lib/common/reflect.ts
+++ b/lib/common/reflect.ts
@@ -27,7 +27,7 @@ function getPropDecorators(directive: any, type: 'Input' | 'Output') {
     (decorators[propName] as any[])
       .filter((decorator: any) => decorator.type.prototype.ngMetadataName === type)
       .forEach((decorator: any) =>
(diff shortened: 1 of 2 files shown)


getsaf:
I pulled this PR down. Need to check on something. May need an additional null check somewhere.

ike18t:
Check out: bac1ca51aad13e9a30615fb6315b61a211cc81e5

Adds emitters to directives.

getsaf:
Added a null check. All good now!

PR #7, opened 2018-03-30
Use angular reflector

getsaf:


lib/common/jit-reflector.ts
```diff
@@ -0,0 +1,34 @@
+import { CompileReflector, ExternalReference } from '@angular/compiler';
```
getsaf: This was based off of Angular's `JitReflector`
https://github.com/angular/angular/blob/c8a1a14b87e5907458e8e87021e47f9796cb3257/packages/platform-browser-dynamic/src/compiler_reflector.ts

getsaf:
Turns out we can't combine `MockDirective` and `MockComponent`, I kind-of like that it worked out that way.

PR #15, opened 2018-08-24
Add documentation for simulated reactive forms events

vespertilian:
Initial documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,12 +9,15 @@ Sure, you could flip a flag on schema errors to make your component dependencies
 
 ## MockComponent(s)
 
-* Mocked component with the same selector
-* Inputs and Outputs with alias support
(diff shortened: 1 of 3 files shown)


vespertilian:
I also want to update this repo:

https://github.com/vespertilian/testing-reactive-forms-third-party-components

to provide a long form example of why we might mock this way, using Angular Materials as the sample code we are mocking, which we could link to later. 

Having some trouble building ng-mocks against the version of angular I am using 6.1.0 vs 6.0.0 in ng-mocks, going to see if when you publish this it's just some weird `link ng-mocks` yarn / npm bug.

ike18t:
Looks like an issue with npm link linking at the root. I just published 6.2 and it shouldn't have that problem.

PR #20, opened 2018-10-02
Support Angular component query decorators

rafehi:
Added the ability to use ViewChild/ViewChildren etc with MockComponent. Also ensured that methods of the component are defined in the mock component, to allow spying. Created ChildComponent to test this functionality.


ike18t:
Great job!

PR #22, opened 2018-10-17
feat(MockDirective): added the ability to use ViewChild/ViewChildren …

kenborge:
…etc with MockDirective

Also ensured that methods of the component are defined in the mock component, to allow spying.

I did the same changes as https://github.com/ike18t/ng-mocks/pull/20 to MockDirectives to be able to use ContentChild and ViewChild with MockedDirectives as well


ike18t:
Thanks @kenborge !

I just released your changes and cut a minor. v6.3.0

PR #34, opened 2018-12-06
Support mock of ModuleWithProviders

Michael-Gusev:
There’re some NgModules with providers like
 * `RouterModule.forRoot`
 * `RouterModule.forChild`
 * `TinymceModule.withConfig`

They return `ModuleWithProviders` instead of `NgModule` and it requires special handling.


Michael-Gusev:
@ike18t, I found one interesting case which wasn't covered by MockModule.
Please review.

ike18t:
Great work! I will cut a release tonight!

PR #35, opened 2018-12-14
Multi MockModule supprt

satanTime:
* Cache for mocked modules
* Cache for mocked pipes
* MockModule can mock modules with the same imports
* MockModule can mock AppModule
* Provider has correct 'provide' property


satanTime:
Hi @ike18t, might you take a look?

lib/mock-module/mock-module.spec.ts
```diff
@@ -44,3 +58,60 @@ describe('MockModule', () => {
...
+
+  it('should be imported correctly', () => {
+    expect(fixture.componentInstance).toEqual(jasmine.any(SameImportsComponent));
+    expect(fixture.nativeElement.innerText).toEqual('same imports');
+  });
+});
+
+// TODO> Doesn't work because ParentModule doesn't export anything.
+// TODO> Basically it's feature of ng-mocks to export declarations of mocked modules.
+// describe('RealModule', () => {
```
satanTime: I would suggest to keep this test so people can understand the difference between mocked module and real one.

lib/mock-module/mock-module.ts
```diff
@@ -1,72 +1,168 @@
...
+  if (ngModuleProviders) {
+    return {
+      ngModule: moduleMockPointer,
+      providers: flatten(ngModuleProviders).map(mockProvider),
+    };
+  }
+  return moduleMockPointer;
+}
+
+const NEVER_MOCK: Array<Type<NgModule>> = [CommonModule, BrowserModule, BrowserAnimationsModule];
```
satanTime: Skipping `BrowserModule`, `BrowserAnimationsModule` allows us to mock root `AppModule`.

satanTime:
Hi @ike18t.  I thinkI found critical issue with this change.
Because everything is cached - spys and other modifications on cached Directives survive between tests and we can get unpredictable assertions later due to missed resets.
I think we need to find a way to listen to events of TestBet to reset cache on every run.

satanTime:
Ha, I guess I'm wrong. I'll investigate deeper and let you know.

satanTime:
Yeah, false alarm.

satanTime:
Understood. So the issue is that only top level module exports everything.
In that case any nested module doesn't export private imports and declarations.
when we have test which requires nested module to be top level and it was already cached - we don't get only original exports instead of everything and it cases the issue.
Working on fix.

ike18t:
Hey @satanTime . 

Yeah that was something I always meant to separate.  Having something like a MockExternalModule and a MockModule (or making it an argument to MockModule) to basically let it know whether it should mock imports vs exports.  

Another thought I had was to allow MockModule take an arg of components not to mock so that you can easily setup your testbed with mocking the module your component under test lives in, but specifying not to mock the component under test.

Unfortunately due to work, wife, and kid my thoughts just stay thoughts.  :-P

satanTime:
Hi there!

Totally agree with ideas and understand with lack of time, the same on my side :)

I'll try my best to commit to your project.
My dream will come true when there will be way to do something like `MockRender('<app-testing-component [test]="somethingVisible"></app-testing-component>')` to avoid small components inside of tests.

I've added commit here to export everything in every module as it was before, please review when you have time https://github.com/ike18t/ng-mocks/pull/37.

Thanks and have a good day!

satanTime:
A sec - I'll add test for the case.

ike18t:
Your dream may already be a reality. A friend of mine wrote the angular testing framework https://www.npmjs.com/package/shallow-render which uses ng-mocks under the hood.

getsaf:
> Your dream may already be a reality.

[Shallow-render](https://github.com/getsaf/shallow-render), making dreams reality ®

satanTime:
Hi. I saw it and tried last week.
Unfortunately, It doesn’t mock components like ‘ng-mocks’ and fails Mocking in my projects.

PR #37, opened 2018-12-17
Avoid of replacement of Application Initializer

satanTime:
* we can mock RouterModule
* we can mock AppModule with imported RouterModule
* mockProvider respects multi flag (avoids Failed: The ANALYZE_FOR_ENTRY_COMPONENTS token only supports ‘multi = true’)


satanTime:
Hi @ike18t, might you review?

PR #40, opened 2018-12-24
Way to mock and to test mocked structural directives

satanTime:


satanTime:
Hi @ike18t, glad to see you're alive. The PR includes also some side optimisations, mostly unit tests. I can create single PR which includes only implementation, or maybe to use patch https://patch-diff.githubusercontent.com/raw/ike18t/ng-mocks/pull/40.patch on current master to have better understanding.

PR #41, opened 2019-01-12
Add MockProvider function

Shinruchan:
Add functions to mock providers/services. Even due to some limitations, I believe, it could prove to be useful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -320,6 +320,56 @@ describe('MockRender', () => {
 });
 ```
 
+## MockProvider
+
+Provides simple way to providers/services when including them in the testing module. 
+
+By default provider will be mocked with an empty method, but if you provide factory method for mocking, all methods will be mocked with that setting. This is useful for example when we want to mock service methods with something like `jasmine.createSpy` or `jest.fn()`.
+
+### Usage Example
+```typescript
+describe('MockProvider', () => {
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [
+        MockProvider(RealService)
+      ]
+    });
+
+    const service = TestBed.get(RealService);
+  });
+});
+```
+
+With factory method:
+```typescript
+describe('MockProvider', () => {
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [
+        MockProvider(RealService, (methodName) => jasmine.createSpy(methodName))
+      ]
+    });
+  });
+
+  it('call service', () => {
+    const service = TestBed.get(RealService);
+
+    // Assuming that component calls the service:
+    expect(service.callMethod).toHaveBeenCalled();
+  });
+});
+```
+
+### Limitations
+
+Current limitations of MockProvider are not being able to mock:
+- properties
+- arrow functions (defined as properties)
+
+
 ## Find an issue or have a request?
 Report it as an issue or submit a PR.  I'm open to contributions.
 
--- a/index.ts
+++ b/index.ts
@@ -5,3 +5,4 @@ export * from './lib/mock-directive';
 export * from './lib/mock-render';
 export * from './lib/mock-module';
 export * from './lib/mock-pipe';
+export * from './lib/mock-provider';
--- a/lib/mock-provider/index.ts
+++ b/lib/mock-provider/index.ts
@@ -1,0 +1,1 @@
+export { MockProvider, MockProviders, IProvider } from './mock-provider';
--- a/lib/mock-provider/mock-provider.spec.ts
+++ b/lib/mock-provider/mock-provider.spec.ts
@@ -1,0 +1,63 @@
+import { async, TestBed } from '@angular/core/testing';
+import { MockProvider } from './mock-provider';
+
+// tslint:disable:max-classes-per-file
+// tslint:disable:prefer-function-over-method
+// tslint:disable:no-unbound-method
+class SampleProvider {
+  method1(): string {
+    return 'test';
+  }
+
+  method2(): void {}
+}
+
+class SampleProvider2 {
+  method1(): void {}
+  method2(): void {}
+}
+
+class SampleProvider3 {
+  method1(): string {
+    return 'test';
+  }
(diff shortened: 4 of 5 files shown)


README.md
```diff
@@ -320,6 +320,56 @@ describe('MockRender', () => {
...
+```
+
+With factory method:
+```typescript
+describe('MockProvider', () => {
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [
+        MockProvider(RealService, (methodName) => jasmine.createSpy(methodName))
```
MartinNuc: Maybe this could happen automatically? Or you need to pass factory method for spy because when one uses JEST the way the spy is created differs? Makes sense :-)

README.md
```diff
@@ -320,6 +320,56 @@ describe('MockRender', () => {
...
+    // Assuming that component calls the service:
+    expect(service.callMethod).toHaveBeenCalled();
+  });
+});
+```
+
+### Limitations
+
+Current limitations of MockProvider are not being able to mock:
+- properties
```
MartinNuc: Actually this would be really useful. We use [spectator](https://github.com/NetanelBasal/spectator) which has `mockProvider` and I added support for mocking properties here https://github.com/NetanelBasal/spectator/pull/58

it works like this:

```
mockProvider(ActivatedRoute, {
    url: activatedRouteUrl,
    snapshot: {}
})
```

`url` and `snapshot` are properties defined on the mock. Other methods are automatically mocked with `jasmine.Spy`.

README.md
Shinruchan: Indeed, to support different testing frameworks this makes sense.

README.md
Shinruchan: I couldn't find a way to read properties from class definition on runtime. But adding a list of them, could be done, indeed.

satanTime:
Hi guys. Please excuse me to interrupt you, but did you check https://github.com/ike18t/ts-mockery or https://github.com/jjherscheid/ts-mocks for example?
There're a lot of libraries to mock simple objects (providers) in typescript or javascript, not sure it requires angular for that.
This ng-mocks library mocks Angular staff, when providers are pure javascript variables.

satanTime:
Think about it in a way: when we mock some Module / Component / Directive / Pipe, they don't use any provider anymore due to mocked behavior, so there's no reason to mock providers.

Usually we need to mock provider when we want to check how real component uses it, it means we want to set behavior of mocked provider and one of them is what to return, that means we can't simply mock some provider, we should also specify its behavior, what brings us back to what we already have in any framework to mock typescript / javascript types such as I mentioned above.

Shinruchan:
I haven't checked those yet. Thanks for sharing. 

The ones that I have checked didn't provide any method of auto-mocking methods, which comes in really handy when one has to use mocked class multiple times.

My thinking behind it was, that if one uses Angular, it's simpler to use one additional library for mocking than multiple. 

But you do have a valid point, so if you guys think this is not feature this package needs, we can discard it.

satanTime:
Let's discuss that with @ike18t, he's owner and maintainer :) 

My 2 cents would to be extend library like https://github.com/ike18t/ts-mockery to get behavior you need and to update readme here with how to mock providers with `ts-mockery`, explain why we have it outside of current project, so people know solution how to mock any part of their application and don't need to google more.

ike18t:
Hi guys!

I agree with @satanTime as far as the responsibility of this library being an Angular template object library.  When we get into providers it gets a little more interesting as far as spy library support and ensuring type-safety.  That is why I was a contributor on ts-mocks and created ts-mockery as a separate lib, I don't want to impose my mocking library biases.

What I can do is add a section to the README like @satanTime suggested to explain how to use ts-mockery for mocking Angular providers.

In the meantime here is a rough example of how to use ts-mockery in conjunction with testbed.

```typescript
class Value {
  getValue: 'hi';
}

class ValueService {
  function foo(arg: string) {
    return arg;
  }

  bizz: 'not buzz';
  doNotCareButWillGetCalled: () => 'whatevs';
  notGoingToGetMocked: () => 'whatevs again';
  value: Value;
}

import { Mock } from 'ts-mockery'

// Within testbed setup
{ 
  provide: ValueService, 
  useValue: Mock.of<ValueService>({ 
    foo: () => 'var', // will also get spied upon
    doNotCareButWillGetCalled: Mock.noop, // don't care about the return but want a spy
    bizz: 'buzz' , 
    value: { getValue: 'not the original value' } // completely type-safe even though it is a nested type and can be a partial
  }) 
}
```

You can mock a partial implementation and can do it recursively if you have nested objects rather than having to nest mocks.  Functions are automatically spied upon if you configure what spy lib 
you are using, it supports jasmine and jest.  One thing to note is that when you create your testbed Angular will make a copy of your providers.  That being said, when you want to modify them you have to pull them out of the testbed in your test and extend it.

```typescript
it('does some stuff', () => {
  const mockedValueService = TestBed.get(ValueService);
  Mock.extend(mockedValueService).with({ bizz: 'still not buzz' });
});
```

Let me know if I forgot to address anything in particular or if there is anything else I can do to help.

Shinruchan:
I guess that settles it then, closing pull request.

MartinNuc:
Thanks for the link. I checked the lib briefly and according to [docs](https://github.com/ike18t/ts-mockery/blob/master/README.md#to-mock-a-function-you-do-not-care-about) it looks like that spies are created only when using Mock.noop. Does it mean we have to spy on every used method manually? I would expect that mocked class has all methods spied.

ike18t:
Any method that you mock, whether it is with a custom implementation or a Mock.noop, will have a spy attached to it.

I just release a minor of ts-mockery which now includes a Mock.all<T> so that you can have an object that is just spies. It works by having a proxy that will create the spies on demand.

Give that a shot and let me know what you think.

Thanks!

MartinNuc:
mockProvider in spectator creates spy automatically for all methods. I am really looking forward to trying out Mock.all! Thank you

Issue #42, opened 2019-01-25
Issue with Animations Library


BuddyLReno:
I'm building a project using Ionic v3 and I'm trying to write tests for some code in my Tabs page.  To do this I have to mock the page components. 

```
TestBed.configureTestingModule({
      declarations: [
        TabsPage,
        MiniPlayerStubComponent,
        MockComponent(ShowsPage)
      ],
```

When I do this, I started getting the error: `Cannot find module '@angular/animations'`.  I don't even use that but all my googling suggested I needed to import it. I did that and now when I run my test, I get this issue:

`TypeError: Cannot read property 'annotations' of undefined`.  

```
at MockDirectiveResolver.Object.<anonymous>.DirectiveResolver.resolve (node_modules/@angular/compiler/@angular/compiler.es5.js:14349:56)
      at MockDirectiveResolver.Object.<anonymous>.MockDirectiveResolver.resolve (packages/compiler/testing/src/directive_resolver_mock.ts:49:1)
      at Object.MockComponent (node_modules/ng-mocks/lib/mock-component/mock-component.ts:24:9)
```

This happens when the mock component tries to mock my ShowsPage.  It looks like it happens when it tries to reflect the class.  I'm not sure how to get past the error at this point, especially because this animations thing is not even something I use. I didn't have to import it until I needed to try mocking this page for testing.

Any help is super appreciated!

satanTime:
I’ll take a look on week. Seems caused by my changes.

satanTime:
Hi @BuddyLReno, could you try latest version of the package?

ike18t:
I'm going to go ahead and close this issue since we haven't heard back. @BuddyLReno feel free to reopen it if your issue is not resolved.

PR #43, opened 2019-01-27
#42: Issue with Animations Library

satanTime:
Added more flexible way to detect system provides, which we shouldn't mock / replace.

### DIFF
--- a/lib/mock-module/mock-module.ts
+++ b/lib/mock-module/mock-module.ts
@@ -1,7 +1,5 @@
 import { CommonModule } from '@angular/common';
 import { ModuleWithProviders, NgModule, Provider, Type } from '@angular/core';
-import { BrowserModule } from '@angular/platform-browser';
-import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { MockOf } from '../common';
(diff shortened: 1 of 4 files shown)


lib/mock-module/mock-module.ts
```diff
@@ -21,15 +19,30 @@ const mockProvider = (provider: any): Provider | undefined => {
...
+  if (typeof provide === 'function' && provide.name === 'RendererFactory2') {
+    return undefined;
+  }
+
+  // BrowserModule injects own providers, which doesn't allow mocks due to own logic.
+  // We have to avoid any injection of those providers to mock everything properly.
+  if (typeof provide === 'function' && provide.name === 'ApplicationInitStatus') {
+    return undefined;
+  }
+  if (typeof provide === 'function' && provide.name === 'DomRendererFactory2') {
```
ike18t: Do you mind squashing the conditionals into one?
```suggestion
  if (typeof provide === 'function' && ['ApplicationInitStatus', 'DomRendererFactory2', ' RendererFactory2'].includes(provide.name) {
```

satanTime: Updated

PR #45, opened 2019-02-15
Feature/mock context with directives

satanTime:
I found interesting case, when empty context triggers failure while render.

Template: `<div *directive=“values; let value;”>{{ value.nested }}</div>`.
when we render it without context we get an error that undefined doesn’t have `nested` property.
It breaks any tests with such variables and due to this limitation we can’t render custom directive with context automatically. The only solution I found is to make them always hidden by default.

Another thing I found: actually there’s a way to check if we have bindings or not, so we could render any structural direcive, but the problem is that all templates like: `<ng-template [directive]>shouldn’t be rendered</template>`, when directive isn’t structural, but attribute, will be rendered when the directive is mocked, because there’s no explicit way to detect is directive structural or not. Any directive on ng-template basically behaves like structural.

So I came to default behavior of structural directives - if we don’t call render inside - it’s not rendered. And we do the same now, we don’t render them, but we can find them by MockHelper and render to do assertions on child elements.


satanTime:
Hi @ike18t, please review.

satanTime:
Ping @ike18t.

Issue #46, opened 2019-03-13
Creating component wrappers


ivi-hamiti:
When testing onpush component to simulate change detection you need to wrap the component with another one and most of the cases have the same input/outputs. So i propose to provide a method, called WrapComponent which mocks a component with the same inputs and outputs of the original one and also wire up the outputs. The template should contain the component we want to wrap and would bind all properties from the wrapper to the inner component.

satanTime:
Hi,

it's here, please check `MockRender`:
```typescript
const fixture = MockRender(
      `
        <tested (trigger)="myListener1($event)" [value1]="myParam1" value2="check">
          <ng-template #header>
            something as ng-template
          </ng-template>
          something as ng-content
        </tested>
      `,
      {
        myListener1: spy,
        myParam1: 'something1',
      }
    );
```
also you can always subscribe to outputs directly, behavior will stay the same as you do that in template.

ivi-hamiti:
yes this makes sense, but effort wise it will be the same effort as to create a component as a class

```
const session = new MockSession();
const wrapperType = WrapComponent(ItemComponent);

urlHelper = session.useMock(UrlHelper);

TestBed.configureTestingModule({
	declarations: [
		wrapperType,
		ItemComponent,
		MockComponent(ItemChildComponent),
	],
	providers: [session.providers()],
});

wrapper = TestBed.createComponent(wrapperType);
```

and the wrapped component template would be something like

`<item [prop1]="prop1" [prop2]="prop2"></item>`

where all these prop can be accessed like

wrapper.componentInstance.prop1....
wrapper.componentInstance.prop2....

as the wrapper will be `ComponentFixture<ItemComponent>` even though it is a wrapper

satanTime:
Exact case of what fixture of MockRender returns, actually it’s default behavior of TestBed itself. Not sure why you need one more wrapper around mocked component, you can access its inputs and outputs as you have on your example.

ivi-hamiti:
In the MockRender case you have to maintain the template rendered manually, if the component inputs or outputs change, the dev will have to change all the things and not always it is type safe. 

Also to clarify the wrapper will wrap the original component, it should not be mocked as i am supposed to test the OnPush behaviour. In the previous example i am testing the `ItemComponent `and i am mocking child components `ItemChildComponent`, but as it is OnPush i am wrapping it around so test ngOnChanges behaviour. So the WrapComponent call will create the component type whill will help me.

So dom wise:
```
[...]

PR #47, opened 2019-03-14
#46: Way to pass Component to MockRender instead of template.

satanTime:


ike18t:
We literally just added this same thing to shallow render a week ago. Great minds think alike. :-)

https://github.com/getsaf/shallow-render/pull/93

PR #55, opened 2019-06-26
fix: es2015 class declaration method mocking

jsantha:
Angular 8 switches TS target to `es2015` and primary bundle contains ES6 class declaration and because of that method mocking in `MockDirective` and `MockComponent` no longer works.

The problem is that native JS class methods are not enumerable and thus `Object.keys` cannot be used. `Object.getOwnPropertyNames` should be used instead to enumerate non enumerable properties.

PR contains changes needed to make mocking work in es2015 environment and are backwards compatible.


field123:
When will this be included in the latest npm version?

ike18t:
Standard version auto bumped minor for some reason but 8.1.0 is released.

Issue #56, opened 2019-06-28
Add value to MockComponent


privettoli:
Expected behaviour:
Values written into MockComponent using `writeValue` should be stored inside custom components for verification of binding in unit tests.

satanTime:
Hi,

are you sure you shouldn't assert call of writeValue with proper arguments?

apapenko:
@satanTime, this was written as a state test, I don't see value in checking writeValue call because the end result is what I'm interested in, not how it's achieved. Feel free to add any other asserts or tests if that's part of the pattern you follow. I just want it to be in release so that tests from parent component that verify custom components value from form is possible (check formControl bindings)

privettoli:
@satanTime, this was written as a state test, I don't see value in checking writeValue call because the end result is what I'm interested in, not how it's achieved. Feel free to add any other asserts or tests if that's part of the pattern you follow. I just want it to be in release so that tests from parent component that verify custom components value from form is possible (check formControl bindings)

satanTime:
Could you provide code of exact test case you try to solve?

privettoli:
@satanTime, it's the one that's in the PR

satanTime:
then there's no reason to change code in the library, you can assert value by `expect(component.writeValue).toHaveBeenCalledWith(expectedValue)`. The idea of mocks is to isolate the behavior of the object you want to replace the other objects by mocks that simulate the behavior of the real objects. And how to simulate the behavior should be defined by developer, not by library. So if you want to set `value` property when `writeValue` has been called you should use `callFake` to set behavior you need to achieve your goals.

privettoli:
Really, that makes perfect sense. :)
Consider the issue to be closed then.

PR #57, opened 2019-06-28
#56 Add value to MockComponent

privettoli:


lib/mock-component/mock-component.ts
```diff
@@ -141,7 +143,9 @@ export function MockComponent<TComponent>(component: Type<TComponent>, metaData?
       this.__simulateTouch = fn;
     }
 
-    writeValue = (value: any) => {};
+    writeValue = (value: any) => {
```
satanTime: the idea of mocking is that you assert call of writeValue with proper args, `ControlValueAccessor` doesn't require to have `value` property.

PR #65, opened 2019-10-25
added MockService helper

tbprojects:
Hi! Thanks for implementing such a helpful mocking tools! I used it on many projects.
The only one thing missing is ability to mock Angular services. I know there is ts-mockery package, but I thought about implementing `MockService` helper, that would be even easier for typical uses. With that solution the ng-mocks would be a complete solution :)

Consider such a component with injected dependency

```typescript
@Component({...})
class SomeComponent {
  constructor(private service: SomeService) {}

  ngOnInit() {
    this.service.getItems().subscribe(...)
  }
}
```

It could be tested with using `MockService(...)` helper

```typescript
beforeEach(() => {
    declarations: [SomeComponent],
    providers: [MockService(ItemsService)]
  });
  
  const service = TestBed.get(ItemsService);
  spyOn(service, 'getItems').and.returnValue(of(['item 1', 'item 2']));

  fixture = TestBed.createComponent(SomeComponent);
  fixture.detectChanges();
})
```

Spying  `getItems` method in separate instruction gives more flexibility in case we would like testing the failing scenario, example:

```typescript
  spyOn(service, 'getItems').and.returnValue(throwError(new Error('failed to fetch items')));
```

What do you think about this :)?
Best wishes, Tomasz

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,6 +450,70 @@ returns first found attribute or structural directive which belongs to current e
 `MockHelper.findDirectives(fixture.debugElement, Directive)`
 returns all found attribute or structural directives which belong to current element and all its child.
 
+## MockService
+
+* extends the original service with optional passing of it's dependencies
+* assumes that you will use `spyOn(..)` for stubbing real methods
+
+### Usage Example
+```typescript
+@Injectable()
+export class NumbersService {
+  constructor(private httpClient: HttpClient) {}
+
+  getNumbers() {
+    return this.httpClient.get<number[]>('/numbers');
+  }
+}
+
+describe('NumbersService', () => {
+  let service: NumberService;
+  let http: HttpClient;
+
+  beforeEach(() => {
+    TestBed.configureTestingModule({
+      providers: [
+        NumbersService,
+        MockService(HttpClient),
+      ]
+    });
+
+    service = TestBed.get(NumbersService);
+    http = TestBed.get(HttpClient)
+  });
+
+  describe('getNumbers', () => {
+    it('should fetch numbers', () => {
+      spyOn(http, 'get').and.returnValue(of([1]));
+   
+      service.getNumbers().subscribe((result) => {
+        expect(result).toEqual([1]);
+      });
+    });   
+  });
+});
+````
+
+## MockedServiceInstance
+
+* useful for getting mocked service instance
+
+### Usage Example
+```typescript
+describe('NumbersService', () => {
+  let service: NumberService;
+  let http: HttpClient;
+
+  beforeEach(() => {
+    http = MockedServiceInstance(HttpClient);
+    service = new NumberService(http);  
+  });
+
+  // ...
+
+});
+````
+
 ## Other examples of tests
 More detailed examples can be found in
 [e2e](https://github.com/ike18t/ng-mocks/tree/master/e2e)
--- a/lib/mock-service/mock-service.spec.ts
+++ b/lib/mock-service/mock-service.spec.ts
@@ -1,0 +1,49 @@
+import { HttpClient } from '@angular/common/http';
+import { TestBed } from '@angular/core/testing';
+import { of } from 'rxjs';
+import { MockedServiceInstance, MockService } from './mock-service';
+import { NumbersService } from './test-services/numbers.service';
+import { StringsService } from './test-services/strings.service';
+
+describe('MockService', () => {
+  describe('when service does not require deps in constructor', () => {
+    beforeEach(() => {
+      TestBed.configureTestingModule({
+        providers: [
+          MockService(NumbersService),
+        ]
+      });
+    });
+
+    it('should allow to mock service', () => {
+      const service: NumbersService = TestBed.get(NumbersService);
+      spyOn(service, 'getNumbers').and.returnValue(of([1]));
+
+      service.getNumbers().subscribe((result) => {
+        expect(result).toEqual([1]);
+      });
+    });
+  });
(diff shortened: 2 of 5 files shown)


LethalsLoaded:
This is really well made, hope to see this merged soon.

gs-akhan:
HI, Please consider merging this.

ike18t:
Maybe I'm not understanding the value add. It appears to just be shorthand for:

`{ provide: ItemService, useValue: {} }`

In the other mock helpers (mockComponent, etc.) there is actual interrogation of angular to find out information about the angular specific object to be mocked via angular reflection so that a mock of the angular object is created. 

I will say that I typically try not to impose my bias on jasmine spies vs type-safe mocking libs but not providing such a helper keeps ng-mocks out of that convo and lets the user make their own decision.

While I am here I would like to point out that jasmine spies are not type-safe and there are plenty of libs out that that give you the benefit of type-safety as well as spy setup. I wrote ts-mockery (but there are other out there) a while back and will use it in the following example.

Screen Shot 2020-01-20 at 11 43 59 AM

You can see that the jasmine spies are cool with returning whatever, no type-safety.

Below is an example of how the included example could be accomplished with a mocking lib:

```typescript
// This lib ensures type-safety
import { Mock } from 'ts-mockery';

beforeEach(() => {
    declarations: [SomeComponent],
    providers: [
      { provider: ItemsService , 
        useValue: Mock.of<ItemsService>({ getItems: () => of([1, 2, 3]) }) }
    ]
  });
})
```

TL;DR  I don't see the added benefit of MockService and ng-mocks is for mocking angular specific objects.

tbprojects:
Thanks @ike18t for your comment :)

Actually `MockService` function produces class with parameter-less constructor by extending the base class that we want to mock - not just an empty object. Thanks to that:
* we cut the injection of further dependencies of the base class
* we have the service instance with all the methods of the base class, that we can mock later on

I didn't provide specific way of mocking instance methods - in my experience it is usually handled after the service instance creation as I may want to test different scenarios. Using `spyOn` is just an example. It could be even mocked by redefining function, ex. `service .getItems = () => of(['item 1', 'item 2'])`
 
I have to agree that jasmine spies are not type-safe. I took a closer look at ts-mockery's `extend` method, which seems to be nice alternative here.

However I still believe that `ng-mocks` would be more complete if `MockService` or similar method is available and developer don't have to look for another library :)

getsaf:
Sorry for the long post, just putting my thoughts down here.

I'm not sure the MockService function qualifies as a "mocking helper" as it does not mock anything. It seems like the goal is to skip out on passing constructor arguments to classes and "hope" that it doesn't negatively affect your test.

According to your comments and [the code](https://github.com/ike18t/ng-mocks/pull/65/files#diff-dc66fc80abac76607379103b87b07b6dR17-R26), it simply extends the service you wish to mock and instantiates the extended class while passing in required dependencies to the parent. All this wrapping serves to do is obfuscate the new-ing up of the actual service with the original constructor and I'm just not clear what benefit that yields.

Using the actual service would lead to brittle tests as you would need to concern yourself with the dependencies and private implementation details of the service-to-be-mocked. You seem to have understood this while coding considering [this error text](https://github.com/ike18t/ng-mocks/pull/65/files#diff-dc66fc80abac76607379103b87b07b6dR6-R8). Dependencies and private implementation details (eg: constructors) are things you should not need to concern yourself with when using mocks. The whole point is to mock them out right?

Also a nit-pick is that the MockService function is capitalized which makes it look like a class. When I see capitalized things in code, I expect to either new it up (eg: new MockService(...)) or use some static property from it (eg: Mock.service()).

Trying to be constructive here so I'll play it out in code.

Say I have a FooService that depends on a BarService which depends on a BazService.

I could do this with MockService:
```typescript
providers: [
  MockService(FooService, [MockService(BarService, [MockService(BazService)])])
]
```
The end result here is that I get three *actual* service instances.

Under the covers, MockService is doing something like this:
```typescript
providers: [
  {provide: FooService, useValue: new FooService(new BarService(new BazService()))}
]
```

Which is functionally identical to:
```typescript
providers: [FooService, BarService, BazService]
```

If my intentions were to integration test, why would I not use the third example? For that matter what does any of this have to do with mocking when the real services are being used?

If my intentions are to unit test, I would forego all the above options and use something like ts-mocks or ts-mockery to actually mock my service.

Using actual mocks seems pretty straight forward and IMO reads more "honestly" in that we are actually using mocks. The next dev that comes along will have no trouble understanding what's happening here.

ts-mockery style:
```typescript
providers: [
  {provide: FooService, useValue: Mock.of<FooService>()}
]
```

ng-mocks style:
```typescript
providers: [
  {provide: FooService, useValue: new Mock<FooService>()}
]
```

Again, sorry for the long post. Hope it makes sense.

tbprojects:
@getsaf thanks for your comment :)

I guess that in 90% of cases you wouldn't have to provide further dependencies to the `MockService(FooService)` if you a not using them directly in FooService constructor.

But you're right, it might be misleading that this helper actually does not mock service methods. Probably I will just use ts-mockery for mocking services. 

Just a final thought: maybe such a `MockService(FooService, args)` method could just return `{provide: FooService, useValue: Mock.of<FooService>(args)}` object in order to have a similar mocking experience as it in other ng-mocks helpers? It would use ts-mockery under the hood if it's optionally installed - I think it could be done with dynamic import.

Feel free to close that pr if there is nothing to be added in the conversation :)

PR #66, opened 2019-11-12
[add] jasmine spy support for all mocked methods

hakobpogh:
Hi guys. I've made some changes that I find to be helpful for me and my team. I'm not sure if you have an issue for this but this is just changing to spy on every method by default instead of putting a dummy version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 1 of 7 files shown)


willemsendennis:
We use Jest, so I rather not have the extra dependency on jasmine. This is what I liked about this project, it's test framework independent.

ike18t:
Hi @hakobpogh ,

First off thank you for your PR to improve on ng-mocks. @willemsendennis has a good point about keeping this lib test framework agnostic. That has always been my goal with this lib and like @willemsendennis I am using jest as well.

Issue #67, opened 2019-11-18
MockRender fixture.detectChanges() call


kekel87:
Hey, we encounter an issue when we using MockRender.
We want to check stuff in `ngAfterViewInit` of tested child component.
But in [MockRender](https://github.com/ike18t/ng-mocks/blob/452818b52ce19d60bbe200fdddb8a14a84303f15/lib/mock-render/mock-render.ts#L60) you call `fixture.detectChanges();`, it's impossible to spy what's going on inside.

It's possible to remove this or make it parametrizable ?
Thanks.

aymeric-duchein:
Hi! I have a similar problem. I'm trying to spy something done inside the constructor of my component.
The issue here is that spies are reset by the `TestBed.configureTestingModule` call. But to achieve what we want, we have to add our spies before the `TestBed.createComponent` call.

A solution to this problem would be to change the mockRender function signature to something like 
(the name of the parameter may be improved 😉 )

```ts
function MockRenderX<MComponent, TComponent extends {[key: string]: any}>(
  template: string | Type<MComponent>,
  params?: TComponent,
  initSpies?: () => void
)
```

and change the end of the function like this: 

```ts

  // Injection of our template.
  TestBed.configureTestingModule({
    declarations: [component],
  });

  if (initSpies) {
    initSpies();
  }

  const fixture = TestBed.createComponent(component);
  fixture.detectChanges();
  return fixture;
```

I will gladly create a PR if this idea suits you.

ike18t:
I added an arg to MockRender to allow more control of when detect changes fires by allowing the user to tell it not to fire. This will allow the user to trigger it themselves. Let me know what you think.

kekel87:
Hi, thank you for your responsiveness. I left you a comment on PR.

kekel87:
I think it would be better if it was false by default.

satanTime:
The fix has been merged and will be released tomorrow.

satanTime:
Please check 9.3.0

PR #71, opened 2020-01-22
feat(mock-render): option to detectChanges or not

ike18t:
#67 Accept an arg to allow control of detectChanges.

### DIFF
--- a/lib/mock-render/mock-render.ts
+++ b/lib/mock-render/mock-render.ts
@@ -5,7 +5,8 @@ import { directiveResolver } from '../common/reflect';
 
 function MockRender<MComponent, TComponent extends {[key: string]: any}>(
   template: string | Type<MComponent>,
-  params?: TComponent
+  params?: TComponent,
+  detectChanges = true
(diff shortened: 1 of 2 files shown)


lib/mock-render/mock-render.ts
```diff
@@ -5,7 +5,8 @@ import { directiveResolver } from '../common/reflect';
 
 function MockRender<MComponent, TComponent extends {[key: string]: any}>(
   template: string | Type<MComponent>,
-  params?: TComponent
+  params?: TComponent,
+  detectChanges = true
```
satanTime: maybe as an object in case if we need to add more flags later?

ike18t: I had that though too...but couldn't think of other options. My thought was that it was easy enough to write an overload/union if we were to pivot to an object.

satanTime: agree

ike18t: Welcome back BTW :-D

PR #83, opened 2020-02-22
Run tests on all versions + Docker for local testing

satanTime:
if you want to run tests locally - simply run `docker-compose up` container has node 10 and chrome browser. It's needed if your local node version is 12+, because some libs can't be compiled successfully on it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ before_script:
     - "sudo chmod 4755 /opt/google/chrome/chrome-sandbox"
 before_install:
   - npm cache clean --force
-script: npm run lint && npm run test
+script: npm run test:angular-versions
 matrix:
   notifications:
     email:
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,2 @@
+./dist
+./node_modules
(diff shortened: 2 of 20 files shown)


satanTime:
Wait, there's issue with puppeteer

satanTime:
fixed, unfortunately, it has to be a dev dependency to avoid redownloading during all versions run.

.travis.yml
```diff
@@ -10,7 +10,7 @@ before_script:
     - "sudo chmod 4755 /opt/google/chrome/chrome-sandbox"
 before_install:
   - npm cache clean --force
-script: npm run lint && npm run test
+script: npm run test:angular-versions
```
ike18t: This all looks great! Do you mind putting the lint verification back in place? It is nice to have for the PR check.

satanTime: it's here, but deep inside. for every version it executes `npm run build:all` which is `npm run lint && npm run test && npm run build`

satanTime: and thanks for the invitation!

ike18t: Thanks for all of your help!

PR #94, opened 2020-03-12
fix: respect of all parents

### LINKED ISSUE #69: Add support for mocking inherited members
It looks like the `MockComponent` method does not setup methods from inherited methods. The code for setting up mock methods for both [MockDirective](https://github.com/ike18t/ng-mocks/blob/master/lib/mock-directive/mock-directive.ts#L73) and [MockComponent](https://github.com/ike18t/ng-mocks/blob/master/lib/mock-component/mock-component.ts#L96) only gets methods from the class's prototype but doesn't traverse up the prototype chain to handle inheritance.

For example, Angular Material's [MatSidenav component](https://github.com/angular/components/blob/master/src/material/sidenav/sidenav.ts#L79) extends the [MatDrawer component](https://github.com/angular/components/blob/master/src/material/sidenav/drawer.ts#L140) and methods defined on MatDrawer are not setup when mocking MatSidenav. This can cause some confusion when some methods on a type are setup and others are not.

[Here's a small stackblitz that reproduces the error](https://stackblitz.com/edit/angular-scjuj2)

satanTime:
Closes #69

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
+<a name="x.x.x"></a>
+# x.x.x (xxx-xx-xx)
+
+
+### Bug Fixes
+
+* **69:** Respect of all parents classes and their methods in mocks
+
+
 <a name="9.1.0"></a>
 # 9.1.0 (2020-03-21)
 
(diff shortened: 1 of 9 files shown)


satanTime:
Hi @ike18t, I've updated PR avoiding usage of the MockBuilder part.

satanTime:
Hi @ike18t, one sec, let me solve the conflict.

ike18t:
@satanTime I merged another of your PRs and it caused a merge conflict in this one.

satanTime:
aha, there's an intersection. this one has been fixed. once it has been merged I can update the MockBuilder's PR.

PR #98, opened 2020-03-28
Prettier, commit lint, commitizen, husky, tslint upgrade

ike18t:


lib/common/jit-reflector.ts
```diff
@@ -12,14 +12,10 @@ export class JitReflector implements CompileReflector {
     return this._reflectionCapabilities.annotations(typeOrFunc);
   }
 
-  componentModuleUrl(type: any, cmpMetadata: Component): string { /* tslint:disable-line prefer-function-over-method */
-    return `./${stringify(type)}`;
-  }
+  componentModuleUrl = (type: unknown) => `./${stringify(type)}`;
```
satanTime: as far as I remember `unknown` doesn't work on older TS versions. But it is awesome!

satanTime:
I would vote for trailing comma.

lib/common/jit-reflector.ts
ike18t: Good catch!

ike18t:
> I would vote for trailing comma.

I hate "dangles" but you and @getsaf would have me outvoted 2 to 1.

satanTime:
I like less diff when someone adds one more item somewhere.

.prettierrc.json
```diff
@@ -0,0 +1,8 @@
+{
+  "bracketSpacing": true,
+  "semi": true,
+  "printWidth": 120,
+  "singleQuote": true,
+  "arrowParens": "avoid",
+  "trailingComma": "es5"
```
ike18t: @satanTime @getsaf because I'm a team player.

PR #99, opened 2020-03-28
fix: never mock ApplicationModule

satanTime:
It has sensetive providers such as:
* Application Initializer
* ApplicationInitStatus
* localeId


ike18t:
10.0.0-alpha.1

PR #100, opened 2020-03-28
feat: support to inject a library-related service mocker

### LINKED ISSUE #87: Add support to inject a library-related service mocker
The idea is to provide support `ts-mockery` as peerDependency and to allow user to mock everything `ng-mocks` mock with a specified library.

For example adding in a setup script of karma or testing framework
```typescript
import 'ng-mocks/mockery/jasmine';
```
or
```typescript
import 'ng-mocks/mockery/jest';
```

and in that case we use specified library to mock methods instead of setting `() => {}` everywhere.

### LINKED ISSUE #103: better spy on getters / setters
with `MockService` we need to set spy on getters and setters correctly because if there is a readonly property it is tough to set its value in the test.

satanTime:
Closes #87
Closes #103

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,8 +439,55 @@ returns attribute or structural directive which belongs to current element.
 `MockHelper.findDirective(fixture.debugElement, Directive)` -
 returns first found attribute or structural directive which belongs to current element or any child.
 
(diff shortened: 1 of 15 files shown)


satanTime:
Oh, I need to upgrade git for the husky.

PR #106, opened 2020-04-07
feat: MockHelper with find, findAll and fetch

satanTime:
Should be rebased and merged after https://github.com/ike18t/ng-mocks/pull/104


satanTime:
Hi @ike18t, I hope this finds you well, might you take a look at the PRs? Can't wait to use it in my projects.

PR #141, opened 2020-06-17
Update project to build to es5 by default

LukeArthurs:
Having the default build to es5 keeps full support for all browsers. IE11 for example which ng-mocks does not work on at the moment since version 9.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -3,6 +3,7 @@
     "lib": ["es2016", "dom"],
     "module": "commonjs",
     "target": "es5",
+    "downlevelIteration": true,
     "noImplicitAny": true,
     "moduleResolution": "node",
     "emitDecoratorMetadata": true,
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "lib": ["es2016", "dom"],
     "module": "commonjs",
-    "target": "es2015",
+    "target": "es5",
```
LukeArthurs, line 5:
```suggestion
    "target": "es5",
    "downlevelIteration": true,
```

satanTime:
Hi, thanks for the report, might you check and fix failed tests and add setup for travis to run tests in IE11 too?

LukeArthurs:
Yes will do! I was hoping it would work in one go but I'll check the tests and update Travis after work today.

satanTime:
After investigation to support es5 by default I found that currently it's impossible to implement due to the next issue: https://stackoverflow.com/a/51860850/13112018.

> Since ES6 classes should be only called with new, NativeBar.call results in error.

That adds limitation to ng-mocks to stay with higher standard.

satanTime:
I've added a separate entry point for es6, should work for IE11.

PR #149, opened 2020-06-25
fix: update NgModuleWithProviders interface

vytautas-petrikas:
This should resolve the issue with Angular 10 https://github.com/ike18t/ng-mocks/issues/148.
However, I am not sure if versions prior to 9 will be compatible...

Either we need to remove the dependency on ModuleWithProviders from angular or to start documenting which ng-mocks versions should be used with which angular version.

Please let me know if anything needs to be adjusted.


satanTime:
Hi. Thanks for the PR. Please add a new interface MockNgModule... that fits all angular versions and we need to add angular 10 to e2e tests too.

vytautas-petrikas:
@satanTime To my understanding I just need to change
```
export interface NgModuleWithProviders<T = any> extends ModuleWithProviders {
  ngModule: Type<T>;
  providers?: Provider[];
}
```
to
```
export interface NgModuleWithProviders<T = any> {
  ngModule: Type<T>;
  providers?: Provider[];
}
```
And of course remove the unused import from @angular/core.

Please clarify if thats what you meant by **Please add a new interface MockNgModule... that fits all angular version**.

P.S. My windows decided to update will add base for e2e tests after update is finished.

satanTime:
Yep. Makes sense

satanTime:
For angular 10. please check in code :a9, the same things should be done for :a10.

vytautas-petrikas:
At the moment only angular10-ivy-true tests are failing, but I will be only able to debug further tomorrow in the evening.

satanTime:
Hi. Thanks for the info. I’m on vacation until Tuesday and won’t be available either.

No rush then. Next week we can solve it.

vytautas-petrikas:
All the tests succeed locally. 
I Have implemented a testBedInjector constant which checks whether TestBed.inject or TestBed.get should be used.
```
const testBedInjector = TestBed['inject'] ? TestBed['inject'] : TestBed['get'];
```
Currently, I am using a string to check the availability of these methods as older versions did not have the **inject** method available.

I think it would be best to create a utility function or constant that can be imported into tests, but I am not sure where would be the best location to place it...

Any guidance on how to fix testBedInjector implementation to succeed on Travis would be appreciated.

satanTime:
first of all thanks for the effort, I checked it yesterday and due to braking changes in ts 3.9 (especially TS2339) to add support of angular 10 requires more changes than a removal of generic, I'll overtake this task.

lib/common/Mock.spec.ts
```diff
@@ -155,12 +155,12 @@ describe('Mock prototype', () => {
 
   it('should get all things mocked and in the same time respect prototype', () => {
     const mockDef = MockComponent(CustomComponent);
-    const mock = new mockDef();
+    const mock = new mockDef() as any;
     expect(mock).toEqual(jasmine.any(CustomComponent));
 
     // checking that it was processed through Mock
-    expect(mock.__ngMocksMock as any).toBe(true);
-    expect(mock.__ngMocksMockControlValueAccessor as any).toBe(true);
```
satanTime, line 163: unfortunately, it won't work like that and should be properly fixed to support TS 3.9. Due to required number of changes in architecture I have to overtake this task.

vytautas-petrikas:
@satanTime I just pushed one last change. I will no longer push new changes on this. Let me know if you need further assistance :)

satanTime:
@vytautas-petrikas, thanks again, might you squash all commits together? I'll try to use your branch as a base one for the full fix.

vytautas-petrikas:
I have squashed the commits

satanTime:
Hi again,

thanks for the effort, unfortunately, with all changes it required a lot of investigation and it was easier to provide a separate PR to fix all of them.

PR #249, opened 2020-12-09
fix(#248): handling null and undefined in declarations

### LINKED ISSUE #248: Cannot read property 'ngModule' of undefined
I have the following setup: I like to test the ProfileOverviewComponent and create mocks for all declarables in its module, which is the ProfileOverviewLazyLoadedModule.

```
 beforeEach(async(() => {
        const metadata = ngMocks.guts(ProfileOverviewComponent, ProfileOverviewLazyLoadedModule);
        metadata.providers.push({ provide: Store, useValue: storeMock });
        TestBed.configureTestingModule(metadata).compileComponents();
    }));
```

But now the test shows the following error: Failed: Cannot read property 'ngModule' of undefined
image

The ProfileOverviewLazyLoadedModule imports around 15 other modules, which have declarables on their own (that need to be mocked in this case). I think some module lookup t read the metadata of the module might be incorrect?

satanTime:
closes #248

### DIFF
--- a/lib/common/func.get-provider.ts
+++ b/lib/common/func.get-provider.ts
@@ -1,0 +1,3 @@
+export default (provider: any): any => {
+  return provider && typeof provider === 'object' && provider.provide ? provider.provide : provider;
+};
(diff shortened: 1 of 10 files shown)


satanTime:
[v11.2.3](https://github.com/ike18t/ng-mocks/releases/tag/v11.2.3) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #260, opened 2020-12-15
build: deps

satanTime:


satanTime:
Hi @ike18t, I've added a bot to update dependencies. But looks like the bot doesn't sign commits and they aren't accepted here. Might you take a look if it's possible to remove this validation or to add an exception for the bot only?

rarkins:
Run the official app if you want signed commits, or use your own signing

PR #318, opened 2021-03-20
feat(#314): ngMocks.formatText

### LINKED ISSUE #314: formatText
There is [`ngMocks.formatHtml`](https://ng-mocks.sudo.eu/api/ngMocks/formatHtml), and I've seen that in `jest` `innerText` doesn't work. At least, for me its value was always empty.

Therefore, to add a helper to get only formatted text makes sense.

But, responsibility how to find related `debugElements` and `debugNodes`, I would suggest to keep outside of this function.

Own helpers can be like that

```typescript
const text = selector => ngMocks.formatText(ngMocks.find(selector));

expect(text('div span')).toEqual('Text 1');
```

```typescript
const allText = selector => ngMocks.findAll(selector)
  .map(ngMocks.formatText);

expect(allText('div span')).toEqual(['Text 1', 'Text 2', 'Text 3']);
```

```typescript
const allTextTrim = selector => ngMocks.findAll(selector)
  .map(ngMocks.formatText)
  .map(v => v.trim());

expect(allText('div span')).toEqual(['Text 1', 'Text 2', 'Text 3']);
```

Idea behind own helper function is that someone may want to lowercase / uppercase text or remove all spaces etc. And to cover all possible cases in `ng-mocks` can end up with lots of helpers which might be complicated to maintain.

satanTime:
closes #314


satanTime:
[v11.10.0](https://github.com/ike18t/ng-mocks/releases/tag/v11.10.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

Issue #377, opened 2021-04-09
.keep(ReactiveFormsModule) causes FormBuilder to malfunction


garrappachc:
Let's say I have a decently simple component that uses the `FormBuilder`:

```ts
@Component({
  selector: 'app-form',
  templateUrl: './form.component.html',
  styleUrls: ['./form.component.css']
})
export class FormComponent {

  form = this.fb.group({
    name: [],
  });

  constructor(
    private fb: FormBuilder,
  ) { }

}
```

```html
<form [formGroup]="form">
  <input type="text" formGroupName="name">
</form>
```

When I test it with ngMocks, like that:

```ts
describe('FormComponent with ngMocks', () => {
  let component: FormComponent;
  let fixture: MockedComponentFixture<FormComponent>;

  beforeEach(() => MockBuilder(FormComponent).keep(ReactiveFormsModule));

  beforeEach(() => {
    fixture = MockRender(FormComponent);
    component = fixture.point.componentInstance;
    fixture.detectChanges();
  });

  it('should create', () => {
    expect(component).toBeTruthy();
  });
});
```

the injected `FormBuilder` instance spits out undefined value, causing angular to fail the unit test:

    Error: formGroup expects a FormGroup instance. Please pass one in.

               Example:

               
            <div [formGroup]="myGroup">
              <input formControlName="firstName">
            </div>

            In your class:

            this.myGroup = new FormGroup({
               firstName: new FormControl()
            });

This error is thrown because the value of `form` is undefined.

I have uploaded a sample repo that reproduces this error: https://github.com/garrappachc/ng-mocks-reactive-forms-bug
In `form.component.spec.ts` you can see two unit tests - one that works (without using ngMocks) and one that doesn't.

Also, I am not sure when was this error introduced, all I know is that Angular version 11.2.5 was working fine and 11.2.6 started to cause the issue - but Angular itself is fine.

Here's my Angular version bump pr that started failing: https://github.com/tf2pickup-org/client/pull/803

satanTime:
Hi, thanks for the report.

Yep. Sorry. My bad. I implemented some wrong changes around 11.2.6. 

Could update to the latest version of ng-mocks? It shouldn't have the issue.

satanTime:
Ah okay. You mean Angular itself.

Then I'll dig into the issue next days and try to provide a fix ASAP.

satanTime:
For now, I would say, that Angular fixed the issue they had. 

[...]

PR #419, opened 2021-04-16
fix: should include the keep module

Teamop:
not sure if it's related with #377, but the issue here is that for the `ReactiveFormsModule`, if using builder to keep `ReactiveFormsModule` and mock another module which also imports the same module, `ReactiveFormsModule` will not be included in the final ngModule. You can refer to the test here. Not sure my solution is totally right, feel free to use the test here for the bug analysis, thanks

### DIFF
--- a/libs/ng-mocks/src/lib/mock-builder/promise/add-missed-keep-declarations-and-modules.ts
+++ b/libs/ng-mocks/src/lib/mock-builder/promise/add-missed-keep-declarations-and-modules.ts
@@ -8,6 +8,10 @@ import { BuilderData, NgMeta } from './types';
 export default (ngModule: NgMeta, { keepDef, configDef }: BuilderData): void => {
   // Adding missed kept providers to test bed.
   for (const def of mapValues(keepDef)) {
+    if (isNgDef(def, 'm') && !ngModule.imports.includes(def)) {
+      ngModule.imports = ngModule.imports.concat(ngMocksUniverse.cacheDeclarations.get(def));
+    }
+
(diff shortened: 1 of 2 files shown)


satanTime:
Hi @Teamop,

I would say, this is expected behavior. The idea of MockBuilder is that the second parameter is the related module of the declaration for the first parameter. Then `ReactiveFormsModule` is available for the testing declaration.

If you want still to get access to `ReactiveFormsModule` outside of the testing declaration, then you should use [export flag](https://ng-mocks.sudo.eu/api/MockBuilder#export-flag).

```ts
    MockBuilder(FormComponent, TestModule)
       .keep(ReactiveFormsModule, {export: true})
```

Teamop:
@satanTime thanks, yes, after adding the `export` flag, everything works.

satanTime:
@Teamop, that's great.

I would recommend to consider usage of `MockBuilder` like `MockBuilder(SomeDeclaration, ModuleWhichDeclaresIt)`. Then you should have proper check of dependencies in unit tests too.

In this case, for the cases when you need to keep a module, such as `ReactiveFormsModule`, another flag should be used: [dependency](https://ng-mocks.sudo.eu/api/MockBuilder#dependency-flag), if someone deletes `ReactiveFormsModule` from imports of `ModuleWhichDeclaresIt`, then unit tests for `SomeDeclaration` will fail.

```ts
MockBuilder(SomeDeclaration, ModuleWhichDeclaresIt).keep(ReactiveFormsModule, {dependency: true});
```

PR #468, opened 2021-05-01
fix: overrides are always precise if they are functions #455

### LINKED ISSUE #455: Allow `precise` flag on `defaultMock`
Yo ! It's me again ! And here I am with a hacky angular DI use case, that works in *some cases* with ng-mocks.

So, first, the intro !

## The goal

I'm trying to inject something that have both a call signature and properties (yep, a function), but **without using the `@Inject` decorator**.

My interface is something like this:

```ts
interface InjectedFn {
  (): string;
  hello: () => number;
}
```

I've found patterns using [abstract classes](https://www.bennadel.com/blog/3836-using-abstract-classes-as-dependency-injection-tokens-with-providedin-semantics-in-angular-9-1-9.htm) that I tried to mimic by attaching the `Injectable` decorator on the function declaration.

```ts
function InjectedFn(_err: any): void {}
Injectable({
  providedIn: "root",
  useFactory: () => {
    const fn: InjectedFn = jasmine.createSpy("Base") as any;
    fn.hello = jasmine.createSpy("Inner");
    return fn;
  }
})(InjectedFn);
```

Then, usage is like this:

```ts
@Component({ template: "" })
export class TestComponent {
  // Note that there is no need to `provide` `InjectedFn` in your NgModule
  public constructor(public myInjectedFn: InjectedFn) {}
}
```

This works great in real applications. But it seems that auto-mocking behavior of ngMocks clashes with this: I suppose it tries to mock methods by replacing the function altogether, thus loosing the call signature.

## Demo

https://stackblitz.com/edit/ng-mocks-examples-f57jx1?file=app/injected-fn.ts

---

I would understand this could be considered as a `wontfix` since this is quite a hack. Moreover, I'm not certain this will always work.
But this works in Angular, including when using Ivy & AOT.
Yet, I suppose the solution of the issue would be quite simple: as demonstrated in the demo above, using the `precise` flag on `MockBuilder.mock` made it work.

satanTime:
closes #455

### DIFF
--- a/libs/ng-mocks/src/lib/common/core.define-property.ts
+++ b/libs/ng-mocks/src/lib/common/core.define-property.ts
@@ -1,13 +1,10 @@
+import helperMockService from '../mock-service/helper.mock-service';
+
 export default (instance: any, property: keyof any, value: any, enumerable = false) => {
-  // istanbul ignore else
-  if (Object.defineProperty) {
-    Object.defineProperty(instance, property, {
-      configurable: true,
-      enumerable,
-      value,
-      writable: true,
-    });
-  } else {
-    instance[property] = value;
-  }
+  helperMockService.definePropertyDescriptor(instance, property, {
+    configurable: true,
+    enumerable,
+    value,
+    writable: true,
+  });
 };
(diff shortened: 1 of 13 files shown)


satanTime:

PR #469, opened 2021-05-01
fix(mock-builder): overrides mock modules for platform #435

### LINKED ISSUE #435: Clarification re: Type is part of the declaration of 2 modules
Hi, and thanks for a great library! I'm currently evaluating this for one of our Angular applications and I've come across one of the issues documented [here](https://ng-mocks.sudo.eu/troubleshooting/declarations-of-2-modules) where I have the following situation:

- Module `A` and module `B` both import `CommonModule` from `@angular/common`.
- Module `B` depends on module `A`.
- I try to test a component in module `B` where I need to mock out something from `CommonModule` explicitly.

I have set up a StackBlitz showcasing what I try to do here: https://stackblitz.com/edit/angular-ivy-7cnyfm?file=src/app/feature/sample.component.spec.ts

I basically have:

- `SharedModule` which imports `CommonModule`.
- `FeatureModule` which imports `SharedModule` and `CommonModule`
- A `SampleComponent` in `SharedModule` which takes a date input and passes it to the Angular `DatePipe`.

I set this up according to my understanding of the docs I linked earlier like this:

```typescript
MockBuilder(SampleComponent)
  .keep(FeatureModule)
  .mock(SharedModule)
  .mock(DatePipe, value => 'MOCK:' + value);
```

This fails with the following error:

> Error: Type NgClass is part of the declarations of 2 modules: CommonModule and MockOfCommonModule! Please consider moving NgClass to a higher module that imports CommonModule and MockOfCommonModule. You can also create a new NgModule that exports and includes NgClass then import that NgModule in CommonModule and MockOfCommonModule.

...and then repeated for more or less everything else in `CommonModule`.

Am I misunderstanding the docs here? I'd be happy to help clarify if I understood what the problem was, but as far as I'm able to tell this should be how things are intended to work.

This error also occurs if I stop including `SharedModule` in my `FeatureModule` and set the tests up like this, leading me to believe this isn't the same case as in the documentation:

```typescript
MockBuilder(SampleComponent, FeatureModule)
  .mock(DatePipe, value => 'MOCK:' + value);
```

Setting it up like in the documentation with `.keep(FeatureModule)` and `.mock(CommonModule)` also yields the same results.

satanTime:
closes #435

### DIFF
--- a/libs/ng-mocks/src/lib/mock-builder/mock-builder.ts
+++ b/libs/ng-mocks/src/lib/mock-builder/mock-builder.ts
@@ -1,9 +1,11 @@
-import { InjectionToken } from '@angular/core';
-import { MetadataOverride, TestBed, TestBedStatic, TestModuleMetadata } from '@angular/core/testing';
+import { InjectionToken, NgModule } from '@angular/core';
+import { getTestBed, MetadataOverride, TestBed, TestBedStatic, TestModuleMetadata } from '@angular/core/testing';
 
+import coreConfig from '../common/core.config';
 import { flatten, mapEntries } from '../common/core.helpers';
+import coreReflectModuleResolve from '../common/core.reflect.module-resolve';
 import { NG_MOCKS, NG_MOCKS_OVERRIDES } from '../common/core.tokens';
-import { AnyType } from '../common/core.types';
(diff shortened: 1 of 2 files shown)


satanTime:

PR #470, opened 2021-05-02
fix(mock-render): generates tpl only for provided inputs and outputs #434

### LINKED ISSUE #434: Defualt values of component inputs are overridden with MockRender
https://stackblitz.com/edit/ng-mocks-examples-mkp6fd?file=app/app.component.spec.ts

In the provided stackblitz example, I expected the testcase to pass. 

When we use `MockRender(AnyComponent)`, the template is generated as 
`<app-any [input1]=input1 [input2]=input2></app-any>`

This overrides the default values of `input1` and `input2` defined in the component class. 

This problem can be solved by using `MockRender('<app-any></app-any>').` 

I feel it will be better if there is a configurable option for controlling template generation. 

- when the second argument to `MockRender()` is `null/undefined` the input/outputs can be ignored in the template generation.
eg: `MockRender(AnyComponent)` should generate the template as `<app-any></app-any>`

- consider only the params provided in second argument for the template generation. 
eg: `MockRender(AnyComponent, { input1: 'abc'})` should generate as `<app-any [input1]="input1"></app-any>`

satanTime:
closes #434

### DIFF
--- a/examples/MockRender/test.spec.ts
+++ b/examples/MockRender/test.spec.ts
@@ -40,8 +40,8 @@ class DependencyModule {}
 })
 class TestedComponent {
   @Output() public readonly trigger = new EventEmitter();
-  @Input() public value1 = '';
-  @Input() public value2 = '';
+  @Input() public value1 = 'default1';
+  @Input() public value2 = 'default2';
 }
 
 describe('MockRender', () => {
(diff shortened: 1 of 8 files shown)


satanTime:

PR #475, opened 2021-05-03
chore(deps): upgrade to Docusaurus 2 alpha.75

slorber:
Supersed https://github.com/ike18t/ng-mocks/pull/466

See also https://github.com/facebook/docusaurus/issues/4709

You can't expect your monorepo structure to work particularly if some existing packages require Webpack 4 and Docusaurus requires Webpack 5. There can only be one Webpack version at a time in `node_modules/webpack`, and Docusaurus wants Webpack 5.


satanTime:
Thanks for the info.

satanTime:
Implemented this approach in the branch of renovate.

PR #581, opened 2021-05-20
fix(faster): supports directives and components without selectors #576

### LINKED ISSUE #576: Every `it` causes an additional `MockMiddleware...` component to be created
First of all, thank you very much for this wonderful package, it has made my life testing an Angular application with _tons_ of services so much better.

I've encountered a strange issue, that I suspect is user error. But I can't for the life of me figure out what I'm doing wrong here, especially because I've used the exact same pattern in a dozen other test files and everything has worked without issue. Just this random one is having a problem.

I can provide my actual test file if its required, but it contains enough company-private information that I'd prefer not to, at least for now. But this is the general idea:

```ts
describe('', () => {
  ngMocks.faster();

  let component: MyComponent;
  let fixture: MockedComponentFixture<MyComponent, any>;

  beforeAll(() => {
    const aThing = { };

    return MockBuilder(MyComponent, ModuleWithMyComponent)
      .provide(MockProvider(ServiceNotInAModule))
      .provide(MockProvider(ComponentScopedService))
      .mock(MyService, {
        someProp: [],
      })
      .mock(MyOtherService, {
        someProp: { },
        anotherProp: new Wrapper(aThing),
      });
  });

  beforeEach(() => {
    fixture = MockRender(MyComponent, {
      someProp: new FormGroup({ }),
      anotherProp: { },
    });
    component = fixture.point.componentInstance;
  });

  it('test 1', () => {
    expect(true).toBe(true);
    // PASSES!
  });

  it('test 2', () => {
    expect(true).toBe(true);
    /*
     FAILS:
       Error: Template parse errors:
       More than one component matched on this element.
       Make sure that only one component's selector can match a given element.
       Conflicting components: MockMiddlewareMyComponent,MockMiddlewareMyComponent ("[ERROR ->]<ng-mocks-MyComponent [someProp]="someProp" [anotherProp]="anotherProp"></ng-mocks-MyComponent>"): ng:///DynamicTestModule/MockRenderComponent.html@0:0
     */
  });

  it('test 3', () => {
    expect(true).toBe(true);
    /*
     FAILS:
       Error: Template parse errors:
       More than one component matched on this element.
       Make sure that only one component's selector can match a given element.
       Conflicting components: MockMiddlewareMyComponent,MockMiddlewareMyComponent,MockMiddlewareMyComponent ("[ERROR ->]<ng-mocks-MyComponent [someProp]="someProp" [anotherProp]="anotherProp"></ng-mocks-MyComponent>"): ng:///DynamicTestModule/MockRenderComponent.html@0:0
     */
  });
});
```

If you look at the errors, you can see that each subsequent test has another of the same component which is conflicting. If I were to re-order these three `it`s, whichever one is first will always pass, and all the subsequent ones will always fail with an increasing number of duplicate components.

As I said, I've used this _exact_ pattern (i.e. `MockBuilder()` in the `beforeAll` and then `MockRender()` in the `beforeEach`) in a dozen or more test files, so I really can't figure out why this is only a problem for this one.

The only weirdness that I can think of which might potentially be an issue is that `MyComponent` injects `ComponentScopedService` in its constructor, and `ComponentScopedService` injects `MyOtherService` in _its_ constructor, so maybe there's some dependency resolution or order of declaration problems with that?

Any help would be appreciated. Thank you very much!

satanTime:
closes #576

### DIFF
--- a/libs/ng-mocks/src/lib/mock-builder/func.extract-tokens.ts
+++ b/libs/ng-mocks/src/lib/mock-builder/func.extract-tokens.ts
@@ -1,0 +1,32 @@
+import { MetadataOverride } from '@angular/core/testing';
+
+import { flatten } from '../common/core.helpers';
+import { NG_MOCKS, NG_MOCKS_OVERRIDES } from '../common/core.tokens';
+import { AnyType } from '../common/core.types';
+
+export default (
+  providers: any,
+): {
+  mocks?: Map<any, any>;
+  overrides?: Map<AnyType<any>, MetadataOverride<any>>;
+} => {
+  let mocks: Map<any, any> | undefined;
+  let overrides: Map<AnyType<any>, MetadataOverride<any>> | undefined;
+
+  for (const provide of flatten(providers || [])) {
+    if (typeof provide !== 'object') {
+      continue;
+    }
+    if (provide.provide === NG_MOCKS) {
+      mocks = provide.useValue;
+    }
+    if (provide.provide === NG_MOCKS_OVERRIDES) {
+      overrides = provide.useValue;
+    }
+  }
+
+  return {
+    mocks,
+    overrides,
+  };
+};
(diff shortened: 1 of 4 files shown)


satanTime:
[v12.0.1](https://github.com/ike18t/ng-mocks/releases/tag/v12.0.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

Issue #610, opened 2021-05-27
MockInstance: ReferenceError: jasmine is not defined


klizter:
I'm having the follow issue with MockInstance in latest version (12.0.2) in combination with jest 27.0.0@next:

```Summary of all failing tests                                                                                                                                   │Enumerating objects: 13, done.
 FAIL  src/app/eq/summary/summary.component.spec.ts                                                                                              │Counting objects: 100% (13/13), done.
  ● SummaryComponent › should create                                                                                                                           │Delta compression using up to 12 threads
                                                                                                                                                               │Compressing objects: 100% (7/7), done.
    ReferenceError: jasmine is not defined                                                                                                                     │Writing objects: 100% (7/7), 9.37 KiB | 505.00 KiB/s, done.
                                                                                                                                                               │Total 7 (delta 6), reused 0 (delta 0), pack-reused 0
      at Object.install (../libs/ng-mocks/src/lib/common/ng-mocks-stack.ts:61:5)                                                                               │remote:
      at restore (../libs/ng-mocks/src/lib/mock-instance/mock-instance.ts:43:18)

```
package.json:
```
{
  ...
  "dependencies": {
    "@angular/animations": "^12.0.1",
    "@angular/cdk": "^12.0.1",
    "@angular/common": "^12.0.1",
    "@angular/compiler": "^12.0.1",
    "@angular/core": "^12.0.1",
    "@angular/forms": "^12.0.1",
    "@angular/platform-browser": "^12.0.1",
    "@angular/platform-browser-dynamic": "^12.0.1",
    "@angular/router": "^12.0.1",
    "@microsoft/applicationinsights-web": "^2.5.9",
    "@navikt/fnrvalidator": "^1.1.3",
    "@ngxs/router-plugin": "^3.6.1",
    "@ngxs/storage-plugin": "^3.6.1",
    "@ngxs/store": "^3.6.1",
    "@types/hammerjs": "^2.0.36",
    "classlist.js": "^1.1.20150312",
    "date-fns": "^2.16.1",
    "dompurify": "^2.0.15",
    "hammerjs": "^2.0.8",
    "js-cookie": "^2.2.1",
    "ngx-mask": "^11.1.4",
    "ngx-webstorage": "^6.0.0",
    "oidc-client": "^1.10.1",
    "rxjs": "^6.6.7",
    "ts-cacheable": "^1.0.1",
    "tslib": "^2.0.3",
    "uuid": "^8.3.0",
[...]

PR #665, opened 2021-06-05
fix(jest): a fix in advance to listen to jest hooks #610

satanTime:


satanTime:
Unfortunately, this solution doesn't work because of missed node env in browser:

```
./node_modules/jest-util/node_modules/picomatch/lib/utils.js:3:11-26 - Error: Module not found: Error: Can't resolve 'path' in '/ng-mocks/e2e/a-min/node_modules/jest-util/node_modules/picomatch/lib'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
	- add a fallback 'resolve.fallback: { "path": require.resolve("path-browserify") }'
	- install 'path-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
	resolve.fallback: { "path": false }
```

PR #727, opened 2021-06-19
fix(mock-builder): provides globally exported providers from directives and components #623

### LINKED ISSUE #623: Issue in mocking service provided in directive in submodule
Hi,

I'm trying to unit test a component (`TodoComponent`), injecting a service (`ContextService`). The service comes from the providers of a directive (`ContextDirective`) instantiating the component dynamically and providing the service as a "context".

The directive is declared and exported from module `SharedModule` and provides the service:

```typescript
@Directive({
  selector: '[appContext]',
  providers: [ContextService]
})
export class ContextDirective implements OnInit
{
  // directive code
}
```

The component is declared in the `TodoListModule`, which imports the `SharedModule`, and injects the ContextService, which is a private instance for the component decorated with the directive:

```typescript
@Component({
  selector: 'app-todo',
  template: `
    <span class="id">{{ todo.id }}.</span>
    <app-description [done]="todo.done" [description]="todo.description"></app-description>
    <input type="checkbox" [(ngModel)]="todo.done" [ngModelOptions]="{ standalone: true }" />
    <button (click)="remove()">Remove</button>
  `,
  styleUrls: ['./todo.component.scss']
})
export class TodoComponent
{
  constructor(private context: ContextService) { }
  // component code
}
```

The setup I'm using for testing the component is the following (I'm using spectator, but I already tried without it and the outcome is the same):

```typescript
describe('TodoComponent', () =>
{
  let spectator: Spectator<TodoComponent>;
  let component: TodoComponent;
  let ctx: SpyObject<ContextService>;


  const module = MockBuilder(TodoComponent, TodoModule)
    .keep(FormsModule)
    .mock(SharedModule)
    .build();

  const factory = createComponentFactory(m(
    module,
    {
      component: TodoComponent,
      detectChanges: false
    }
  ));

  beforeEach(() =>
  {
    spectator = factory();
    component = spectator.component;
    ctx = spectator.inject(ContextService);
  });

  // actual tests
```

I get the error `NullInjectorError: No provider for ContextService!`, because it seems that the MockBuilder is not mocking the service.

I tried also adding `.mock(ContextService)`, with imho should be sufficient because the component shouldn't know who is the provider of the service, it just needs an instance, but the error persists.

If I remove the `.mock(SharedModule)` and I mock explicitly the `ContextService` it works, but I get a lot of warnings because a subcomponent (`DescriptionComponent`) defined in `SharedModule` is not mocked, and I would like that it and all the other dependencies from `SharedModule` are mocked.

I tried debugging the issue, and it seems that the `.mock(ContextService)` line is ignored because the check in https://github.com/ike18t/ng-mocks/blob/master/libs/ng-mocks/src/lib/mock-builder/promise/add-missed-mock-declarations-and-modules.ts#L14 aborts the mocking process for `ContextService` because it is already present in the `ngMocksUniverse.touches` Set.

Here (https://github.com/DavideCanton/ng-mocks-issue) there is the repo for reproducing the issue, the tests not working are the ones for the `TodoComponent`, and they are already marked with `fdescribe`, so a `npm run test` is sufficient to reproduce the issue.

Am I missing something in configuring the `MockBuilder`?

satanTime:
closes #623

### DIFF
--- a/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.ts
+++ b/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.ts
@@ -8,8 +8,8 @@ import { isNgModuleDefWithProviders } from '../common/func.is-ng-module-def-with
 import ngMocksUniverse from '../common/ng-mocks-universe';
 
 import { MockBuilderStash } from './mock-builder-stash';
-import addMissedKeepDeclarationsAndModules from './promise/add-missed-keep-declarations-and-modules';
-import addMissedMockDeclarationsAndModules from './promise/add-missed-mock-declarations-and-modules';
+import addMissingKeepDeclarationsAndModules from './promise/add-missing-keep-declarations-and-modules';
+import addMissingMockDeclarationsAndModules from './promise/add-missing-mock-declarations-and-modules';
 import addRequestedProviders from './promise/add-requested-providers';
 import createNgMocksOverridesToken from './promise/create-ng-mocks-overrides-token';
 import createNgMocksToken from './promise/create-ng-mocks-token';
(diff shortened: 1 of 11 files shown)


satanTime:
[v12.1.1](https://github.com/ike18t/ng-mocks/releases/tag/v12.1.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #887, opened 2021-07-31
docs(MockRender): information about proxies #839

### LINKED ISSUE #839: Modifying component input on the wrapper modifies the param object supplied to MockRender
### Scenario:
A Component has an Input:

```ts
class AppComponent {
  @Input()
  name: string = 'default';

  constructor() {}
}
```

A param object including a value for this input is supplied to `MockRender`:
```ts
const params = {
  name: 'originalParam'
};
fixture = MockRender(AppComponent, params);
```

The _wrapper_ input is modified:
```ts
fixture.componentInstance.name = 'modifiedOnWrapper'; // not fixture.point.componentInstance
fixture.detectChanges();
```

### Expected result:
Only the wrapper input and component inputs are modified:
```js
expect(fixture.componentInstance.name).toBe('modifiedOnWrapper');
expect(fixture.point.componentInstance.name).toBe('modifiedOnWrapper');
```

### Actual result:
The param object is also modified:
```js
expect(params.name).toBe('originalParam'); // Fails
```

[Example in Stackblitz](https://stackblitz.com/edit/github-zfdnjo-oa8uhj?devtoolsheight=33&file=src/examples/main/test.spec.ts)

### Additional Info:
ng-mocks version: 11.11.2 in my project, 12.3.1 in the StackBlitz example

I expected references to be maintained, otherwise you couldn't provide a spy via the param object and assert on it, but I didn't expect assigning a string literal to the wrapper's input property to also modify the param object.  In the docs, it's also a bit unclear which is preferred, assigning to the wrapper or the component.  Most examples modify the wrapper, but the page for `MockRender` calls it "mostly useless", begging the question why you'd set inputs through it then.

satanTime:
closes #839

### DIFF
--- a/docs/articles/api/MockComponent.md
+++ b/docs/articles/api/MockComponent.md
@@ -27,9 +27,9 @@ The class of a mock component has:
 - the same `selector`
 - the same `@Inputs` and `@Outputs` with alias support
 - templates with pure `<ng-content>` tags to allow transclusion
-- support of `@ContentChild` and `@ContentChildren`
-- support of `ControlValueAccessor`, `Validator` and `AsyncValidator`
-- support of `exportAs`
+- support for `@ContentChild` and `@ContentChildren`
+- support for `ControlValueAccessor`, `Validator` and `AsyncValidator`
+- support for `exportAs`
 
 :::tip
 Information about mocking FormControl, ControlValueAccessor, Validator and AsyncValidator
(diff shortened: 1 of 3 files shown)


satanTime:
[v12.5.0](https://github.com/ike18t/ng-mocks/releases/tag/v12.5.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #925, opened 2021-08-06
Improve MockRenderFactory sample code

jsmart523:
Documenting that `ngMocks.faster()` is required.

### DIFF
--- a/docs/articles/api/MockRender.md
+++ b/docs/articles/api/MockRender.md
@@ -177,21 +177,31 @@ Considering the conditions above, we would need to create a factory once with he
 and then 5 tests should call the factory in order to create fixtures.
 
 ```ts
+import { Input } from '@angular/core';
+import { ngMocks, MockBuilder, MockRenderFactory } from 'ng-mocks';
+
+class MyComponent {
+  @Input() public input1 = 1;
+  @Input() public input2 = 2;
+}
+
 describe('Maximum performance', () => {
   const factory = MockRenderFactory(MyComponent, ['input1', 'input2']);
   
-  ngMocks.faster();
-  beforeAll(() => MockBuilder(MyComponent, MyModule));
+  ngMocks.faster(); // REQUIRED when using MockRenderFactory
+  beforeAll(() => MockBuilder(MyComponent));
   beforeAll(() => factory.configureTestBed());
 
   it('covers one case', () => {
-    const fixture = factory({input1: 1});
-    expect(fixture.point.componentInstance.input1).toEqual(1);
+    const fixture = factory({input1: 11});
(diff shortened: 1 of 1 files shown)


satanTime:
Hi @jsmart523,

thanks for the PR.

I think it is a bug actually. Because it shouldn't require `ngMocks.faster`.

Would be cool if you could investigate why it doesn't work without `ngMocks.faster`, otherwise I'll try to dig into it in 1-2 weeks.

docs/articles/api/MockRender.md
```diff
@@ -177,21 +177,31 @@ Considering the conditions above, we would need to create a factory once with he
...
+  @Input() public input2 = 2;
+}
+
 describe('Maximum performance', () => {
   const factory = MockRenderFactory(MyComponent, ['input1', 'input2']);
   
-  ngMocks.faster();
-  beforeAll(() => MockBuilder(MyComponent, MyModule));
+  ngMocks.faster(); // REQUIRED when using MockRenderFactory
+  beforeAll(() => MockBuilder(MyComponent));
```
satanTime, line 192: ah wait. it had `ngMocks.faster()` already.

Might you explain how you stumbled on it?

jsmart523: I'm in a large developer team with unit tests using only raw angular. Conversion to spectator create much better code (no NO_ERRORS_SCEMA needed, and runs 871% faster even though it mocks at run-time) but I am evaluating this library by itself because I'm confused:  "Why use spectator if spectator tells us to use ng-mocks for mocking components and there's so much functional overlap?!?"

So I'm converting unit tests, and wanted to just convert original to ng-mocks and see how that feels and performs, and *then* try it again with `ngMocks.faster()` and see how the numbers change. So I had intended to add `ngMocks.faster()`, just not yet, then the code didn't work and I was left in newbieland where I didn't know what I didn't know, so I copied your code, tried it, and then was playing with it to see why yours worked and mine didn't!

jsmart523: I suspected it might be a bug, but wasn't sure. For all I know it was mentioned five times elsewhere in the documentation and I missed it. ;-)

Either way, it's a low priority bug, in my peanut-gallery opinion. If you're showing how to optimize things, why would anybody (besides me) not use `ngMocks.faster()` as well?

I'm unfamiliar with the library and have some deadlines, so I'm not going to be digging into the code. Until you or somebody does though, I hereby offer helpful documentation! :-P

satanTime:
Hi, I hope this finds you well.

I'm closing this PR because I wasn't able to reproduce the issue for the factory doesn't show its dependency on faster.

PR #1993, opened 2022-03-06
feat(MockInstance): resets root overrides likewise properties #1256

### LINKED ISSUE #1256: MockProvider does not reset the token value of MockInstance
First of all, thank you for this awesome plugin. We integrated it in our solutions and its working really great! Unfortunately one thing is not working as I expect it, but maybe its not a bug, but i missunderstood it.

I created an angular injection token and mock it with MockProvider, i.e.
```
beforeEach(() => {
        providers: [
            MockProvider(myToken, { test: 1 })
        ],
    });
});
```
Inside an 'it' I override it with another value:
`MockInstance(myToken, () => ({ test: 2 }));`

When the next 'it' runs, I expect the value to be resettet to '{ test: 1 }' again by the MockProvider call in beforeEach, but I still get the value '{ test: 2 }' here.

Am I using this the wrong way or is this a bug of the framework?

satanTime:
close #1256

### DIFF
--- a/libs/ng-mocks/src/lib/common/ng-mocks-universe.ts
+++ b/libs/ng-mocks/src/lib/common/ng-mocks-universe.ts
@@ -54,14 +54,6 @@ ngMocksUniverse.global.set('flags', {
   onTestBedFlushNeed: coreConfig.onTestBedFlushNeed,
 });
 
-ngMocksUniverse.getLocalMocks = () => {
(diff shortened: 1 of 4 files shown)


satanTime:
[v13.1.0](https://github.com/ike18t/ng-mocks/releases/tag/v13.1.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #2054, opened 2022-03-13
fix(core): BrowserAnimationsModule is optional dependency now #1377

### LINKED ISSUE #1377: Library tests fail for projects which don't include @angular/animations module
We have a [library](https://angular.io/guide/creating-libraries#refactoring-parts-of-an-application-into-a-library) where in tests we use `ng-mocks`.  Our project doesn't include `@angular/animations` module as we don't use animations in our app. Our Angular is `12.2.13`

After updating `ng-mocks` from 11.x to latest 12.x (`12.5.0` at the moment of writing), following error started to appear when running library tests (`ng test <library-name>`), preventing tests from being run:

```
./node_modules/@angular/platform-browser/fesm2015/animations.js:9:0-83 - Error: Module not found: Error: Can't resolve '@angular/animations' in '/Users/vzozuliak/projects/dxng/node_modules/@angular/platform-browser/fesm2015'

./node_modules/@angular/platform-browser/fesm2015/animations.js:10:0-227 - Error: Module not found: Error: Can't resolve '@angular/animations/browser' in '/Users/vzozuliak/projects/dxng/node_modules/@angular/platform-browser/fesm2015'
```

With further investigation traced the issue to https://github.com/ike18t/ng-mocks/commit/ca65de3a2747fd1ac69c2dcef47a4896c7aa13a2 which was introduced in `12.1.0`. Double checked and re-assured that version `12.0.2` works well. 

Also, went to dist code of `12.5.0` and commented out this code in `ng-mocks/cjs/lib/mock-module/create-resolvers.js`, and again it proved to work just fine:

https://github.com/ike18t/ng-mocks/blob/17d798e74730fedff3b6af69f5fda70eecb89cc7/libs/ng-mocks/src/lib/mock-module/create-resolvers.ts#L19-L31

My assumption - Angular library builder is slightly different from app builder. I assume it tries to analyze all files for `require`s and to include them in it's own way, and so it tries to include  `@angular/platform-browser/animations` which then tries to include `@angular/animations` and so that fails, but since it all happens out of context of executing the actual file, it's not get caught with that `try/catch`

Workarounds:
- downgrade to `ng-mocks` 12.0.2 (not cool as we stuck with old version)
- write a script that patches `ng-mocks` commenting out the code above (not coll at all - monkey patching should be the last possible resort)

satanTime:
closes #1377

### DIFF
--- a/docs/articles/troubleshooting/browser-animations-module.md
+++ b/docs/articles/troubleshooting/browser-animations-module.md
@@ -4,38 +4,54 @@ description: Information and solutions how to test modules with BrowserAnimation
 sidebar_label: BrowserAnimationsModule
 ---
 
-By default, `ng-mocks` replaces `BrowserAnimationsModule` with `NoopAnimationsModule`.
+By default, Angular recommends replacing `BrowserAnimationsModule` with `NoopAnimationsModule` in unit tests.
 
-However, it can be changed via [`MockBuilder`](../api/MockBuilder.md) or [`ngMocks.guts`](../api/ngMocks/guts.md)
-when `NoopAnimationsModule` isn't a solution.
+In order to do so globally, you can use [`ngMocks.globalReplace`](../api/ngMocks/globalReplace.md):
+
+```ts title="src/test.ts"
+ngMocks.globalReplace(BrowserAnimationsModule, NoopAnimationsModule);
+```
+
+Now, every time `ng-mocks` sees `BrowserAnimationsModule`, it will substitute it with `NoopAnimationsModule`.
 
 ## MockBuilder
 
-```ts
-// No animations at all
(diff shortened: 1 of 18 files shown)


zuzusik:
According to [test results](https://app.circleci.com/pipelines/github/ike18t/ng-mocks/4178/workflows/bb930123-6973-408a-9b90-3463d14ecb29/jobs/157115/parallel-runs/0/steps/0-105)  It seems like this actually perfectly reproduces original issue brought up in https://github.com/ike18t/ng-mocks/issues/1377 😄

satanTime:
[v13.4.0](https://github.com/ike18t/ng-mocks/releases/tag/v13.4.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #2096, opened 2022-03-19
fix(MockInstance): correctly accepts falsy values #2087

### LINKED ISSUE #2087: `MockInstance` on component injected as a provider does not work for falsy values
Hey there ! Hope you are doing well !

In one of my tests, a mocked component is injected as a provider in my actually tested component, and `MockInstance` seems to not work as expected: it does mock only truthy values, not falsy ones.

See the demo here: https://stackblitz.com/edit/github-ytvy85?file=src/test.spec.ts

What is strange is that the behavior is inconsistent:
* either this is not how to test this situation, and thus, all tests should fail
* or this is OK, and falsy values should be mocked too.

I think the problem comes from here: https://github.com/ike18t/ng-mocks/blob/0cfd6dc7f2cb2e95fdff3ca68433460e0a89f308/libs/ng-mocks/src/lib/mock-instance/mock-instance.ts#L182-L184

> *Footnote:* You have links to stackblitz examples. What could be really cool would be to prepare a super trimmed down template for issues reporting, maybe even with both karma & jest presets, and request users to post a repro if possible in your issue template. 

Have a nice day, and thanks again for the great lib ;)

satanTime:
closes #2087

### DIFF
--- a/libs/ng-mocks/src/lib/mock-instance/mock-instance.ts
+++ b/libs/ng-mocks/src/lib/mock-instance/mock-instance.ts
@@ -42,8 +42,8 @@ const parseMockInstanceArgs = (args: any[]): MockInstanceArgs => {
     set.accessor = args[2];
   } else {
     set.value = args[0];
-    if (typeof set.value !== 'function') {
-      set.value = set.value?.init;
+    if (set.value && typeof set.value === 'object') {
+      set.value = set.value.init;
     }
   }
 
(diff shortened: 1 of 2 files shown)


satanTime:
[v13.2.0](https://github.com/ike18t/ng-mocks/releases/tag/v13.2.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #2124, opened 2022-03-26
fix(ngMocks.stubMember): forwarding stub values to point.componentInstance #1165

### LINKED ISSUE #1165: stubMember should respect proxies
if we define stubMember on `fixture.componentInstance`, which is a proxy to `fixture.point.componentInstance`, then the setter should be called instead of the property override.

The setter: https://github.com/ike18t/ng-mocks/blob/75cabd9bf8181355f51ead7be00dc309e90bd174/libs/ng-mocks/src/lib/mock-render/func.install-prop-reader.ts#L20-L29

satanTime:
closes #1165


satanTime:
[v13.3.0](https://github.com/ike18t/ng-mocks/releases/tag/v13.3.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #2202, opened 2022-04-07
fix(core): exporting types in package.json #2198

### LINKED ISSUE #2198: Auto-import from ng-mocks doesn't work (VS Code)
### Description of the bug

Cannot use auto-import in VS Code like with all other packages

E.g. if I use MockBuilder somewhere in my code and it's not defined (typescript says 'cannot find name 'MockBuilder'.ts(2304)'  I expect that pressing 'Quick fix' button will present me an option 'add import from "ng-mocks", that automatically adds `import { MockBuilder } from 'ng-mocks'` line.
Tried it in several projects.

Usually this feature work perfectly for other packages, probably it's something related to the way the exports are done from 'ng-mocks'.

Listing all imports manually each time is kind of annoying.

satanTime:
closes #2198


satanTime:
[v13.4.1](https://github.com/ike18t/ng-mocks/releases/tag/v13.4.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #2650, opened 2022-06-02
fix(MockBuilder): respect extention of classes with different decorators #2646

### LINKED ISSUE #2646: Bug: MockBuilder.mock component that extends a service results in mock that's placed in providers rather than declarations
### Description of the bug

Let's say we have a component that extends a service. Trying to call `MockBuilder.mock(MyComponent).build()` on that component results in the component mocked as a service and placed in the providers. Please see the repro link.
I suspect that similar behavior would be observed if we call `.keep(...)`

Link: https://codesandbox.io/s/cool-swartz-f4nnwz?file=/src/test.spec.ts

The mocked component should end up in `declarations`

satanTime:
closes #2646


satanTime:
[v14.0.0](https://github.com/ike18t/ng-mocks/releases/tag/v14.0.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #2741, opened 2022-06-09
feat(MockBuilder): default params as dependency or export #2647

### LINKED ISSUE #2647: Bug: A module that is explicitly marked to be kept should be kept even if it's imported in another module that is mocked
### Description of the bug

Let's say that my component under test uses some other components that come from 3rd party library. These components are made available through a module that the 3rd party library also provides. Under the hood, the 3rd party library module is using `ReactiveFormsModule`.

Now, lets say that my component also does some forms stuff, so I need to use `ReactiveFormsModule` as well. Therefore I'd want to mock all the components that are coming from the 3rd party library, but keep `ReactiveFormsModule`.
I'd think that I should achieve this behavior via the following:
```ts
const deps = MockBuilder()
      .mock(AwesomeLibraryModule) // the 3rd party library module
      .keep(ReactiveFormsModule)
      .build();
```
This however does not seem to work. The `ReactiveFormsModule` is not found in the `deps.imports`. Please see the repro link

Link: https://codesandbox.io/s/nostalgic-moore-7sezz5?file=/src/test.spec.ts

The `ReactiveFormsModule` should be included in the `deps.imports`.

satanTime:
closes #2647

- [x] throw if a dependency wasn't found
- [x] update docs about form testing
- [x] update migration docs about the breaking change

BREAKING CHANGE: MockBuilder with 2 params marks all chain calls as dependency
BREAKING CHANGE: MockBuilder with 0-1 params marks all chain calls as export

### DIFF
--- a/docs/articles/api/MockBuilder.md
+++ b/docs/articles/api/MockBuilder.md
@@ -11,15 +11,21 @@ but with minimum overhead.
 Usually, we have something simple to test, but time to time, the simplicity is killed by nightmarish dependencies.
 The good thing here is that commonly the dependencies have been declared or imported in the same module, where our
 tested thing is. Therefore, with help of `MockBuilder` we can quite easily define a testing module,
-where **everything in the module will be replaced with their mocks**, except the tested thing: `MockBuilder( TheThing, ItsModule )`.
+where **everything in the module will be replaced with their mocks**, except the tested thing:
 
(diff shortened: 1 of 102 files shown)


satanTime:
[v14.0.0](https://github.com/ike18t/ng-mocks/releases/tag/v14.0.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #2848, opened 2022-06-19
fix(core): provide mjs in package.json #2846

### LINKED ISSUE #2846: Getting failures from jest when trying to use MockComponent
### Description of the bug

When using Angular 14, jest and jest-preset-angular, testing with ESM, it fails when I try to use MockComponent.  In the attached zip you can see a simple barebones project that has a single test file. The package.json uses the nasty node command that seems to be required for this.

I tried putting the mock in the imports as well, instead of the declarations, but it didn't seem to make a difference.

[foo.zip](https://github.com/ike18t/ng-mocks/files/8934971/foo.zip)

satanTime:
closes #2846


satanTime:
[v14.0.1](https://github.com/ike18t/ng-mocks/releases/tag/v14.0.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #3105, opened 2022-07-15
fix(core): detecting and mocking standalone directives correctly #3100

### LINKED ISSUE #3100: Bug: MockBuilder a standalone component that imports a standalone directive, it fails
### Description of the bug

When mocking a standalone directive imported by a standalone component, an error is throw saying that MockOfDirective is not standalone

`Error: The "MockOfTestDirective" directive, imported from "TargetComponent", is not standalone. Did you forget to add the standalone: true flag?`

### An example of the bug

https://stackblitz.com/edit/github-7zh6cj?file=src%2Ftest.spec.ts

Link:

Since the original imported directive (TestDirective) is a standalone: true directive, the mocked version (MockOfTestDirective) should also be a standalone and the MockBuilder should be able to mock it as expected.

Not sure if I am missing anything or doing something wrong, but I did the same test with a standalone Pipe and it worked ok. 
Is this scenario already supported ? 
What I could be doing wrong here ?

Thanks !

satanTime:
closes #3100

### DIFF
--- a/libs/ng-mocks/src/lib/mock-component/mock-component.ts
+++ b/libs/ng-mocks/src/lib/mock-component/mock-component.ts
@@ -198,14 +198,11 @@ coreDefineProperty(ComponentMockBase, 'parameters', [
 
 const decorateClass = (component: Type<any>, mock: Type<any>): void => {
   const meta = coreReflectDirectiveResolve(component);
-  const template = generateTemplate(meta.queries);
-  const mockParams = {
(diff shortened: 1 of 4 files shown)


satanTime:
[v14.1.0](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.1.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #3231, opened 2022-07-31
fix(MockBuilder): respects global configuration for standalone dependencies #3161

### LINKED ISSUE #3161: ngMocks.globalReplace doesn't seem to replace in standalone component
### Description

I apologize that I don't have an easy reproduction added here but the basic issue is the following, when using standalone components (testing with NgMocks v14.1.0) any ngMocks.globalReplace calls don't seem to apply to those components.

So in our codebase we do a global setup that calls something like this:
```javascript
ngMocks.globalReplace(TranslateModule, MockTranslateModule);
```

Then in our code we have a standalone component:
```javascript
@Component({
  select: 'a-component',
  template: '',
  standalong: true,
  imports: [TranslateModule]
})
```

and the test setup for the standalone component
```javascript
  beforeEach(() => {
    return MockBuilder(TheStandaloneComponent);
  });
```

However the test don't work because when the tests run it uses vanilla/automatic module mocks for the TranslateModule instead of using my global replacement.

### Workaround
I can work around it by adding a replace directly in the test like this then it works again:
```javascript
  beforeEach(() => {
    return MockBuilder(TheStandaloneComponent).replace(TranslateModule, MockTranslateModule);
  });
```

satanTime:
closes #3161

MockBuilder will process dependencies of standalone declarations with respect of the next configuration:

- ngMocks.globalKeep
- ngMocks.globalMock
- ngMocks.globalReplace
- ngMocks.globalExclude

### DIFF
--- a/libs/ng-mocks/src/lib/mock-builder/promise/init-keep-def.ts
+++ b/libs/ng-mocks/src/lib/mock-builder/promise/init-keep-def.ts
@@ -3,7 +3,7 @@ import { funcExtractDeps } from '../../common/func.extract-deps';
 import ngMocksUniverse from '../../common/ng-mocks-universe';
 
 export default (keepDef: Set<any>, configDef: Map<any, any>): Set<any> => {
-  const mockDef = new Set<any>();
+  const dependencies = new Set<any>();
   const builtDeclarations = ngMocksUniverse.builtDeclarations;
   const builtProviders = ngMocksUniverse.builtProviders;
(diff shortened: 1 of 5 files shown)


satanTime:
[v14.1.1](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.1.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #3795, opened 2022-10-09
fix(core): exporting internal types #3709

### LINKED ISSUE #3709: Consider making Type<T> public again
We use ng-mocks to create re-usable mock modules so we don't need to repeat our mocking set-up across apps.  This is done using standard Angular CLI library projects and `ng-packagr`.

Since v14, the following interface is no long public:

```typescript
export interface Type<T> extends Function {
  new (...args: any[]): T;
}
```

This means that usage such as the below is no longer possible:

```typescript
export const FooTestingModule = MockModule(FooComponentsModule);
```

Compilation now results in the following error:

`Exported variable 'FooTestingModule ' has or is using name 'Type' from external module "E:/code/Foo/node_modules/ng-mocks/index" but cannot be named.`

Could this interface be made public again or a different approach suggested to satisfy the above requirement?

satanTime:
closes #3709


satanTime:
[v14.2.4](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.2.4) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #3899, opened 2022-10-24
docs: updating use cases for ngMocks.get

Dji75:
Currently `ngMocks.get` description could be understood to only concerns structural directives or services but it can also be used for components or common directives `declarations` word is too vague and can be more precise.

Consequently we may be tempted to use `ngMocks.findInstance` which mention components, directives and so on in the documentation. But in some cases we could have false positive tests (using `ngMocks.findInstance` rather than `ngMocks.get`), so documentation update is needed for `ngMocks.get`.

@satanTime let me know if the doc update looks great or if something else could be made event better.

### DIFF
--- a/docs/articles/api/ngMocks/get.md
+++ b/docs/articles/api/ngMocks/get.md
@@ -3,19 +3,39 @@ title: ngMocks.get
 description: Documentation about ngMocks.get from ng-mocks library
 ---
 
-Returns a declaration, service or token, which can be attribute or structural directives,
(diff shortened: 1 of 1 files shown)


satanTime (approved):

satanTime:
Looks great! Thank you for contributing!

satanTime:
[v14.3.2](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.3.2) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #4502, opened 2022-12-10
feat(MockInstance): ignores undefined properties #4367

### LINKED ISSUE #4367: Feature Request: allow to mock only specific methods

Sometimes I want to mock just one method and leave the rest of the service as is. So, it can run some domain specific logic (or side-effects)

### Proposed solution

Provide a function that has the same declaration as `MockProvider` but modifies service instead of creating a full mock:

```ts
class MyService {
  someMethod() {
       console.log('implementation')
       return of(undefined);
  }

  anotherMethod() {
        return this.someMethod().pipe(
          tap(() => {
              // side effects, changes to service properties or changes in dependencies
          })
        )
  }
}

PatchProvider(MyService, {
  someMethod: jest.fn(() => of(...))
});

// later in tests
const service = TestBed.inject(MyService);

console.log(service.someMethod) // jest.fn()
console.log(service.anotherMethod) // original method
```

Additionally would be nice to ignore `undefined` values in object, i.e., if property is `undefined` just ignore it and do not override service method/property:

```ts
PatchProvider(MyService, {
  someMethod: jest.fn(() => of(...)),
  anotherMethod: undefined // ignored, as if there is no this property at all
});
```

this allows to simplify pattern of using `setup` function and passing props inside:

```ts
function setup(props?: Props) {
       providers: [
           PatchProvider(MyService, {
              someMethod: props?.someMethod
           })
       ]
    });
}

interface Props {
  someMethod?: MyService["someMethod"]
}
```

We use this pattern a lot in our codebase, together with `MockProvider`, so would be nice to have this built-in.

### Additional context

We have implemented this custom function using `APP_INITIALIZER` token:

```ts
function PatchProvider<T extends {}>(
    token: ProviderToken<T>,
    modifications: Partial<T>
): FactoryProvider {
  return {
        provide: APP_INITIALIZER,
        multi: true,
        deps: [token],
        useFactory(service: T) {
            return () =>
                Object.keys(modifications).forEach((key) => {
                    const value = (modifications as Record<string, unknown>)[key];
                    if (value === undefined) return;

                    (service as Record<string, unknown>)[key] = value;
                });
        },
    };
}
```

satanTime:
closes #4367

### DIFF
--- a/libs/ng-mocks/src/lib/mock-helper/mock-helper.stub.ts
+++ b/libs/ng-mocks/src/lib/mock-helper/mock-helper.stub.ts
@@ -19,6 +19,9 @@ export default <T = MockedFunction>(instance: any, override: any, style?: 'get' 
 
   for (const key of Object.getOwnPropertyNames(applyOverrides)) {
     const desc = skipProps.indexOf(key) === -1 ? Object.getOwnPropertyDescriptor(applyOverrides, key) : undefined;
+    if (desc && Object.prototype.hasOwnProperty.call(desc, 'value') && desc.value === undefined) {
+      continue;
+    }
     helperMockService.definePropertyDescriptor(correctInstance, key, desc);
   }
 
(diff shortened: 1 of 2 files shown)


satanTime:
[v14.5.0](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.5.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #4628, opened 2023-01-06
fix(MockBuilder): imports modules with providers on root level #4613

### LINKED ISSUE #4613: Bug: Provider imported by module is deleted if submodule does not provide it
### Description of the bug

Say we have a module X, and a module Y with a dependency on X. In the testing module, X is imported with some providers in the testing module, in module Y it is not. Consequently, the providers should be available in the testing module. But now, say we have a combination of real components and fake components (so that the "magic" in https://github.com/help-me-mom/ng-mocks/blob/35aa7f1ba69aa97fd2e8c0db43d2aa5e1064296d/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts#L214 is applied). Then the provider is not available anymore in v14.5.1, but was in v14.5.0.

### An example of the bug

Link: https://stackblitz.com/edit/angular-ivy-fgudvv?file=src%2Fapp%2Fapp.component.spec.ts

Expected behavior: FooComponent can be created.

Actual behavior: FooComponent cannot be created because no provider for MyService is found.

satanTime:
closes #4613

### DIFF
--- a/libs/ng-mocks/src/lib/mock-builder/promise/init-ng-modules.ts
+++ b/libs/ng-mocks/src/lib/mock-builder/promise/init-ng-modules.ts
@@ -70,6 +70,8 @@ export default (
     if (!config.dependency && config.export && !configInstance?.exported && (isNgDef(def, 'i') || !isNgDef(def))) {
       handleDef(meta, def, defProviders);
       markProviders([def]);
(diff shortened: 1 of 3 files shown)


satanTime:
[v14.5.2](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.5.2) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

Issue #7216, opened 2023-11-09
Bug: ngMocks.findInstance unable to find instance (Angular 17)


GipHub123:
### Description of the bug

It look's like something has changed in recently released Angular 17. 
In some test cases ngMocks.findInstance is unable to find instance (Standalone).
`ngMocks.findInstance(TestComponent);`

   TypeError: Cannot read properties of null (reading 'rootNodes')
          at Ft (node_modules/ng-mocks/index.mjs:1:69900)
          at node_modules/ng-mocks/index.mjs:1:70211
          at Gt (node_modules/ng-mocks/index.mjs:1:70279)
          at Wt (node_modules/ng-mocks/index.mjs:1:70594)
          at Qt (node_modules/ng-mocks/index.mjs:1:71245)
          at Qt (node_modules/ng-mocks/index.mjs:1:71254)
          at Qt (node_modules/ng-mocks/index.mjs:1:71254)
          at Qt (node_modules/ng-mocks/index.mjs:1:71254)
          at Qt (node_modules/ng-mocks/index.mjs:1:71254)
          at Qt (node_modules/ng-mocks/index.mjs:1:71254)

All failing test cases has been setup by using `MockBuilder`.
All test cases (~4000) were fine before updating to Angular 17.

I've mainly used an alternative method to find instances.  
`
  function get(fixture: ComponentFixture<any>, type: Type<any>): any {
    return fixture?.debugElement?.query(By.directive(type)).componentInstance 
  }
`

I replaced `ngMocks.findInstance` with above `get`-function and it fixed the problem. 

### An example of the bug

Link:

### Expected vs actual behavior
`ngMocks.findInstance` find an instance without an error. 

```


Angular CLI: 17.0.0
Node: 20.9.0
Package Manager: npm 10.1.0
OS: win32 x64

Angular: 17.0.1
... animations, common, compiler, compiler-cli, core, forms
... language-service, platform-browser, platform-browser-dynamic
... platform-server, router

Package                                    Version
--------------------------------------------------------------------
@angular-devkit/architect                  0.1700.0 (cli-only)
@angular-devkit/build-angular              17.0.0
@angular-devkit/core                       17.0.0 (cli-only)
@angular-devkit/schematics                 17.0.0
@angular/cdk                               17.0.0
@angular/cli                               17.0.0
@angular/google-maps                       17.0.0
@angular/material                          17.0.0
@angular/material-date-fns-adapter         17.0.0
[...]

PR #8757, opened 2024-04-12
fix: unable to find instance in case of ng17 control flow

andreandersson:
ViewContainerRef actually returns the wrong length in case of Angular 17 Control Flow. By checking that the returned value is not null we should be safe.

Solves #7216


andreandersson:
@satanTime apparently I'm not sure how the test-setup should be here, where should I add the test so that:

a) we get coverage report for it correctly
b) it works for a16 and below

satanTime (approved):

satanTime:
Thank you a lot for the contribution!

satanTime:
[v14.12.2](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.12.2) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #8771, opened 2024-04-16
Remove ViewChild from metadata

### LINKED ISSUE #8634: Bug: Incompatibility of MockComponent with new viewChild signal function
### Description of the bug

Testing standalone components with nested components that use the new viewChild signal feature is currently not possible if the nested component should be mocked.

I'm not sure if ngMocks can do anything about it or if its an angular-core related problem.

### An example of the bug

Link: https://stackblitz.com/edit/github-6eneyd?file=src%2Ftest.spec.ts
I added a "NestedComponent" to the default example and replaced the module-based definitions with a simple standalone definition. 

Using the Signal implementation leads to
`TypeError: Cannot read properties of undefined (reading 'Symbol(SIGNAL)')`
Switching the implementation from Signal to Decorator lets test run fine,

c-harding:
ViewChild will always be null, and so there is no point in mocking it.
viewChild.required throws an error in mocked components without this
change.

Fixes GH-8634


satanTime (changes requested):
Hi there.

Thank you for contribution, but it's a breaking change which affects other angular versions. Should be fixed without breaking anything.

c-harding:
Hi @satanTime, what’s the effect for other versions? Only `__prop_metadata__` is affected, which is not a documented property. Additionally, viewChild would always be falsy for mocked components anyway, because mocked components do not have any of their own elements as children.

satanTime:
ng-mocks uses many undocumented and monkey patched things. 

For a proper PR, you need to provide a proper test which demonstrates the issue and its fix.

Here, I see that mocks which provide own empty views now will return nulls instead of objects, so tests which relied on these empty views will fail despite no change in the code, only due to the update of ng-mocks. 

That's what should be avoided, so what worked before should still work afterwards.

dmitry-stepanenko:
@satanTime signal-based queries have `isSignal: true` property set. WDYT if we would modify the condition as follows
```typescript
const query: Query & { ngMetadataName?: string; isSignal?: boolean } = queries[key];
const isSignalBasedQuery = query.isViewQuery && query.isSignal;
if (!isSignalBasedQuery) {
    final.push([key, query]);
}
```

That way it should not be considered as a regression for any existing cases while still allowing us to move forward

c-harding:
@dmitry-stepanenko that sounds promising! I’ve given you write access to this repo, so you’re welcome to edit the PR directly

dmitry-stepanenko:
awesome, thanks @c-harding. I'll update the branch shortly

kheos31:
Hello! @satanTime does the fix provided by @dmitry-stepanenko in this PR and as he explained in this post https://github.com/help-me-mom/ng-mocks/pull/8771#issuecomment-2128806068 is what you expected as a good fix? 
Otherwise, how could we help to do this PR "acceptable"? 
Thanks in advance

gestj:
Pump ... Sad this is still pending... known issue since now almost a year :/

Francesco-Borzi:
any news?

Watercycle:
For those who really want to use signals right now or are facing the struggles of mixed code, here's a not-fully-vetted patch containing this branch and #11909:
```json
"ng-mocks": "https://github.com/Watercycle/ng-mocks/releases/download/v14.14.0/ng-mocks-0.0.1.tgz",
```

I think the next steps for this branch are... re-merging master? It seems like the tests didn't run last time ([CircleCI](https://app.circleci.com/pipelines/github/help-me-mom/ng-mocks?branch=pull%2F8771)). I'd be surprised if anything significant came up, though.

Watercycle:
I've duplicated this PR over to #12875 with some minor changes to (artificially) make CI happy. I think ideally `tests-e2e` would also report coverage and be merged with the existing `tests` coverage to simplify things for future contributions; but, that's a separate adventure and kinda pending the question of whether it's worth the effort of having every release maintain compatibility across every Angular version.

satanTime:

c-harding:
Wasn’t this already fixed in v14.15.1? 🤔 

https://github.com/help-me-mom/ng-mocks/issues/8634#issuecomment-4002162928

satanTime:
Who knows :) I only hope it's been really fixed

PR #9090, opened 2024-05-30
build: refresh package-lock.json file with latest @angular

matheo:
After some time updating dependencies leaves some old logic or nested dependencies, and from time to time it's good to remove `node_modules` and the `package-lock.json`, and install from scratch.

Any ETA for the v18 release?
Thanks in advance!


satanTime:
Sorry. This PR doesn't make sense.

Issue #1008, opened 2021-08-30
Cover all guards variants


satanTime:
The question came from gitter: https://gitter.im/ng-mocks/community?at=6128e79d4c7be06b7986702b

Hi guys.
I'm trying to test a route resolver and everything seems to be working fine for the successful resolve, but for unsuccessful resolves, I redirect the user to a different route using
```ts
            return from(
                this.router.navigate(['app/maintenance'], { state: { tab } })
            ).pipe(map(() => null));
```
The problem seems to be that when I check the location.path, it has not navigated to the route yet (Error: Expected '/' to equal '/app/maintenance'). I can also see in coverage that the map operator has not been hit at all, which leads me to believe that the router.navigate promise has not resolved yet when the test ends.
I've tried multiple tick, flush, flushMicrotasks calls but I'm not able to get this router navigate to resolve.
Can anyone assist with this?

---

One thing I noticed is that the route has a guard on, and to get the tests to run I had to mock that guard, even though I'm adding `.exclude(NG_MOCKS_GUARDS)`.

Here is my MockBuilder
```ts
return MockBuilder(WidgetResolver, [
    RuntimeModule,
    AppModule,
    MaintenanceModule,
    HomeModule,
    StoreFeaturesModule
])
    .exclude(NG_MOCKS_GUARDS)
    .keep(RouterModule)
    .keep(RouterTestingModule.withRoutes([]))
    .keep(RuntimeComponent)
    .keep(StoreRootModule)
    .mock(WidgetPreviewComponent)
    .mock(MaintenanceComponent)
    .mock(MessageService)
    .mock(MaintenanceGuard)
    .provide(provideMockStore({ initialState }));
```

---

If I don't mock that guard then I get
```
Error: Uncaught (in promise): TypeError: Cannot read property 'pipe' of undefined TypeError: Cannot read property 'pipe' of undefined at MaintenanceGuard.checkForMaintenancePermission (src/app/guards/maintenance.guard.ts:34:51) at MaintenanceGuard.canLoad (src/app/guards/maintenance.guard.ts:22:6)
```

---

Does the `NG_MOCKS_GUARDS` work for canLoadGuards?

satanTime:
[v14.13.1](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.13.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

satanTime:
[...]

PR #9551, opened 2024-07-25
docs: can-load guard #1008

### LINKED ISSUE #9835: Feature Request: CanLoad coverage
https://angular.dev/api/router/CanLoad

windmichael:
I have added a new can-load.spec to verify that `NG_MOCKS_GUARDS` is working also with the canLoad guard. 
In addition I updated the docs to have link this the can-load.spec instaed of the can-match.spec. 

Hope this PR helps. 
Thanks for the good work. I really like using ng-mocks.

fixes #9835

### DIFF
--- a/examples/TestRoutingGuard/can-load.spec.ts
+++ b/examples/TestRoutingGuard/can-load.spec.ts
@@ -1,0 +1,172 @@
+import { Location } from '@angular/common';
+import {
+  Component,
+  inject,
+  Injectable,
+  NgModule,
+  VERSION,
+} from '@angular/core';
+import {
+  CanLoadFn,
+  Router,
+  RouterModule,
+  RouterOutlet,
+} from '@angular/router';
+import { RouterTestingModule } from '@angular/router/testing';
+import { from } from 'rxjs';
+import { mapTo } from 'rxjs/operators';
+
+import {
+  MockBuilder,
+  MockRender,
+  NG_MOCKS_GUARDS,
+  NG_MOCKS_ROOT_PROVIDERS,
+  ngMocks,
+} from 'ng-mocks';
+
+// A simple service simulating login check.
+// It will be replaced with it's mock copy.
+@Injectable()
+class LoginService {
+  public isLoggedIn = false;
+}
+
+// The canLoad guard we want to test.
+const canLoadGuard: CanLoadFn = (route, segments) => {
+  if (route && segments && inject(LoginService).isLoggedIn) {
+    return true;
+  }
+
+  return from(inject(Router).navigate(['/login'])).pipe(mapTo(false));
+};
+
+// Another canLoad guard like in a real world example,
+// which should be removed from testing to avoid side effects on the route.
+const sideEffectCanLoadGuard: CanLoadFn = () => false;
+
+// A simple component pretending to be a login form.
+// It will be replaced with a mock copy.
+@Component({
+  selector: 'login',
+  template: 'login',
+})
+class LoginComponent {}
+
+// A simple component pretending to be a protected dashboard.
+// It will be replaced with a mock copy.
+@Component({
+  selector: 'dashboard',
+  template: 'dashboard',
+})
+class DashboardComponent {}
+
+@NgModule({
+  declarations: [DashboardComponent],
+  imports: [
+    RouterModule.forChild([
+      {
+        path: 'dashboard',
+        component: DashboardComponent,
+      },
+      {
+        path: '**',
+        redirectTo: 'dashboard',
+      },
+    ]),
+  ],
+  exports: [],
+})
+class DashboardModule {}
+
+// Definition of the routing module.
+@NgModule({
+  declarations: [LoginComponent, DashboardComponent],
+  exports: [RouterModule],
+  imports: [
+    RouterModule.forRoot([
+      {
+        component: LoginComponent,
+        path: 'login',
+      },
+      {
+        canLoad: [canLoadGuard, sideEffectCanLoadGuard],
+        path: '',
+        loadChildren: () => DashboardModule,
+      },
+    ]),
+  ],
+  providers: [LoginService],
+})
+class TargetModule {}
+
+describe('TestRoutingGuard:canLoad', () => {
+  // Because we want to test a canLoad guard, it means that we want to
+  // test it's integration with RouterModule.
+  // Therefore, RouterModule and guard should be kept,
+  // and the rest of the module which defines the route can be mocked.
+  // To configure the RouterModule for the test,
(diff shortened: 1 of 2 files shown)


examples/TestRoutingGuard/can-load.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+import {
+  MockBuilder,
+  MockRender,
+  NG_MOCKS_GUARDS,
+  NG_MOCKS_ROOT_PROVIDERS,
+  ngMocks,
+} from 'ng-mocks';
+
+// A simple service simulating login check.
+// It will be replaced with its mock copy.
```
JsantosDK, line 28: // It will be replaced with its mock copy. -> // It will be replaced with it's mock copy.

examples/TestRoutingGuard/can-load.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  return from(inject(Router).navigate(['/login'])).pipe(mapTo(false));
+};
+
+// Another canLoad guard like in a real world example,
+// which should be removed from testing to avoid side effects on the route.
+const sideEffectCanLoadGuard: CanLoadFn = () => false;
+
+// A simple component pretending a login form.
```
JsantosDK, line 47: // A simple component pretending a login form. -> // A simple component pretending **to be** a login form.

examples/TestRoutingGuard/can-load.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+// It will be replaced with a mock copy.
+@Component({
+  selector: 'login',
+  template: 'login',
+})
+class LoginComponent {}
+
+// A simple component pretending a protected dashboard.
```
JsantosDK, line 55: // A simple component pretending a protected dashboard. -> // A simple component pretending **to be** a protected dashboard.

examples/TestRoutingGuard/can-load.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+      },
+    ]),
+  ],
+  providers: [LoginService],
+})
+class TargetModule {}
+
+describe('TestRoutingGuard:canLoad', () => {
+  // Because we want to test a canLoad guard, it means that we want to
+  // test its integration with RouterModule.
```
JsantosDK, line 104: // test its integration with RouterModule. -> //  test it**'**s integration with RouterModule.

examples/TestRoutingGuard/can-load.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+  // Therefore, RouterModule and the guard should be kept,
```
JsantosDK, line 105: // Therefore, RouterModule and the guard should be kept, -> // **Therefore the RouterModule and guard** should be kept,

examples/TestRoutingGuard/can-load.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+        loadChildren: () => DashboardModule,
+      },
+    ]),
+  ],
+  providers: [LoginService],
+})
+class TargetModule {}
+
+describe('TestRoutingGuard:canLoad', () => {
+  // Because we want to test a canLoad guard, it means that we want to
```
JsantosDK, line 103: Update the text to resolve some minor sentence stuff.
  // Because we want to test a canLoad guard, it means that we want to
  // test it's integration with RouterModule.
  // Therefore, RouterModule and guard should be kept,
  // and the rest of the module which defines the route can be mocked.
  // To configure the RouterModule for the test,
  // RouterModule, RouterTestingModule.withRoutes([]), NG_MOCKS_ROOT_PROVIDERS
  // should be specified as the first parameter of MockBuilder (with empty routes).
  // The module with routes and the guard should be specified
  // as the second parameter of MockBuilder.
  // Then `NG_MOCKS_GUARDS` should be excluded to remove all guards,
  // and `canLoadGuard` should be kept to let you test it.

examples/TestRoutingGuard/can-load.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+
+      return;
+    }
+
+    const fixture = MockRender(RouterOutlet, {});
+    const router = ngMocks.get(Router);
+    const location = ngMocks.get(Location);
+
+    // First we need to initialize navigation.
+    if (fixture.ngZone) {
```
JsantosDK, line 139: Starting with Angular now comes with the option to be zoneless.
Does this spec function under this condition?
Could we trigger the navigation in another way, so to not rely on the ngZone?

JsantosDK:
Hello @windmichael, I checked out the PR, for me it seems in line what is already done in the lib.
Pointed out some sentence stuff, if you could correct that would be great.
Also, did you test this with Angular 18? Asking because not sure that the way you trigger the navigation will work in that version.

examples/TestRoutingGuard/can-load.spec.ts
windmichael: You raised a valid point here. 
I would rather consider creating a new issue to check, if everything is still working on a zoneless application. The way how the test is implemented here, is also used in other specs. 
Probably it would make sense to have e2e tests for Angular 18 Zoneless. But as I said therefore a new issue should be created.

JsantosDK: @satanTime 
From looking into the past stuff, I saw that there where two branches at one point because of the different rendering engines.

Do you think this could be an approach here?
Having 2 branches per version, after Angular 18, one for zoneless and another with it?

I'm aware from reading the angular team blogs and github that they want to move away from it, but not sure if it will actually occur, or when. Using the two branches approaches is the safest I can think of.

satanTime (approved):

examples/TestRoutingGuard/can-load.spec.ts
satanTime: it's just a test, and, for now, it's fine that it depends on `zone.js`, because `ng-mocks` examples include this library by default.

It doesn't block usage of `ng-mocks` and `canLoad` in projects without `zone.js`.

satanTime (approved):

satanTime:
[v14.13.1](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.13.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

satanTime:
[v14.13.1](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.13.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #10021, opened 2024-09-14
fix: improves error message (#10020)

### LINKED ISSUE #10020: Bug: Unhelpful error message when MockBuilder .keep() or mock() is used instead of .provide()
### Description of the bug

When accidentally mocking a provider dependency with `MockBuilder` using `.keep()` or `.mock()` instead of `.provide()`, I get the following mysterious and uninformative error message:
```
Error: MockBuilder has found a missing dependency: Object. It means no module provides it. Please, use the "export" flag if you want to add it explicitly. https://ng-mocks.sudo.eu/api/MockBuilder#export-flag
```

### An example of the bug

For example, try providing `ActivatedRoute` with the `.keep()`:
```
  beforeEach(() => {
    return MockBuilder(TargetComponent, TargetModule).keep({
      provide: ActivatedRoute,
      useValue: {
        queryParams: of({ query: '', keys: [] }),
        queryParamMap: of({ query: '', keys: [] }),
      },
    });
  });
```

The test fails with the error: `MockBuilder has found a missing dependency: Object.`

Changing `.keep()` to `.provide()` fixes the problem:
```
  beforeEach(() => {
    return MockBuilder(TargetComponent, TargetModule).provide({
      provide: ActivatedRoute,
      useValue: {
        queryParams: of({ query: '', keys: [] }),
        queryParamMap: of({ query: '', keys: [] }),
      },
    });
  });
```

Link: https://stackblitz.com/edit/github-hzu37y?file=src%2Ftest.spec.ts

#### Expected behavior

Test fails to run with a helpful error message, such as:

```
Error: A provider object has been incorrectly passed to the MockerBuilder.keep() method. Did you mean to use MockerBuilder.provide()?
```

#### Actual behavior

Test fails to run with the following error message, which is not helpful (e.g., adding `keep(Object, {export: true}` still fails).

```
Error: MockBuilder has found a missing dependency: Object. It means no module provides it. Please, use the "export" flag if you want to add it explicitly. https://ng-mocks.sudo.eu/api/MockBuilder#export-flag
```

ajlouie:
Fixes #10020: Throws error with a more helpful message if a provider object is passed to `keep` or `mock`.


satanTime (approved):

satanTime:
Hi @ajlouie, thank you for help!

Issue #10487, opened 2024-11-16
No updates until Jan 2025


satanTime:
I guess you all noticed no active development of the library for several years. 

It's going to change from January 2025. 

Thanks to all who was patient and hasn't lost trust in ng-mocks.

JPtenBerge:
Not to add to any stress, but will Angular 19 get a relatively quick version support release before that time by any chance?

evtk:
Thanks for the update @satanTime. I still use ng-mocks on daily basis as it abstracts so much of the unit test fuzz. 

I came here after watching the updates for NG19 and noticing how there is getting more and more attention to implementing signals. The simple fact is that ng-mocks currently prevents me from starting to use the signal input api for instance instead of the regular Input decorator. 

So really looking forward to updates on that part!

JPtenBerge:
@evtk ng-mocks hasn't prevented me from using signal inputs.

```ts
export class InterviewStepsComponent {
    step = input.required<Step>();
```

```ts
fixture = TestBed.createComponent(InterviewStepsComponent);
sut = fixture.componentInstance;
fixture.componentRef.setInput('step', currentStep);
```

I'm not using `MockRender()` though, is that what you're using?

GipHub123:
@evtk 
I can confirm that input signals work with the current version of ngMocks. However, as far as I know, certain features, such as `ContentChildren `and `ViewChild`, are not yet supported.
 https://github.com/help-me-mom/ng-mocks/issues/8895

evtk:
> @evtk ng-mocks hasn't prevented me from using signal inputs. ng-mocks simply doesn't provide any help in that area.
> 
> ```ts
> export class MyStepComponent {
>     step = input.required<Step>();
> ```
> 
> ```ts
> beforeEach(() => {
>     TestBed.configureTestingModule({
>         providers: [MockProvider(...), MockProvider(...)],
>         imports: [MyStepComponent],
>     });
> 
>     fixture = TestBed.createComponent(MyStepComponent);
>     sut = fixture.componentInstance;
>     fixture.componentRef.setInput('step', currentStep);
> });
> ```
> 
> I'm not using `MockRender()` though, is that what you're using?

Yeah definitely went for being `Pro` and using `MockRender` everywhere in my tests :|

JPtenBerge:
@satanTime Understood, thanks for the response.

dmitry-stepanenko:
[...]

PR #10583, opened 2024-11-26
fix(a19): respecting default standalone flag since angular 19

re-alam:
Added standalone false as default for mock-render and updated isStandalone method to handle angular 19+ default standalone behavior for components.

### DIFF
--- a/libs/ng-mocks/src/index.ts
+++ b/libs/ng-mocks/src/index.ts
@@ -4,6 +4,7 @@ import './lib/common/ng-mocks-stack';
 import './lib/common/ng-mocks-global-overrides';
 
 export * from './lib/common/core.tokens';
+export * from './lib/common/func.is-standalone';
 
 export { getTestBedInjection, getInjection } from './lib/common/core.helpers';
 
--- a/libs/ng-mocks/src/lib/common/func.is-standalone.ts
+++ b/libs/ng-mocks/src/lib/common/func.is-standalone.ts
@@ -1,3 +1,5 @@
+import { VERSION } from '@angular/core';
+
 import collectDeclarations from '../resolve/collect-declarations';
 
 import { getNgType } from './func.get-ng-type';
@@ -11,5 +13,11 @@ export function isStandalone(declaration: any): boolean {
     return false;
   }
 
-  return collectDeclarations(declaration)[type].standalone === true;
+  // Handle Angular 19+ default standalone behavior
+  const declarations = collectDeclarations(declaration);
+  if (Number(VERSION.major) >= 19 && type !== 'NgModule' && declarations[type].standalone === undefined) {
+    return true;
+  }
+
+  return declarations[type].standalone === true;
 }
--- a/libs/ng-mocks/src/lib/mock-render/func.create-wrapper.ts
+++ b/libs/ng-mocks/src/lib/mock-render/func.create-wrapper.ts
@@ -140,6 +140,7 @@ export default (
     selector: 'mock-render',
     template: mockTemplate,
     viewProviders: flags.viewProviders,
+    standalone: false,
   };
 
   ctor = generateWrapperComponent({ ...meta, bindings, options });
--- a/tests/func.is-standalone.spec.ts
+++ b/tests/func.is-standalone.spec.ts
@@ -1,0 +1,30 @@
+import { Component, VERSION } from '@angular/core';
+
+import { isStandalone } from 'ng-mocks';
+
+@Component({
+  selector: 'standalone',
+  template: `<div>
+    <h1>Angular 19 standalone</h1>
+  </div>`,
+})
+class StandaloneComponent {}
+
+describe('func.is-standalone', () => {
+  describe('Angular 19+ specific tests', () => {
+    let originalMajor: string = VERSION.major;
+    const setVersionMajor = (major: number) => {
+      originalMajor = VERSION.major;
+      Object.assign(VERSION, { ...VERSION, major: major.toString() });
+    };
+
+    afterEach(() => {
+      Object.assign(VERSION, { ...VERSION, major: originalMajor });
+    });
+
+    it('should return true when standalone is undefined', () => {
+      setVersionMajor(19);
+      expect(isStandalone(StandaloneComponent)).toBeTruthy();
+    });
+  });
+});


libs/ng-mocks/src/lib/common/func.is-standalone.ts
```diff
@@ -11,5 +13,11 @@ export function isStandalone(declaration: any): boolean {
     return false;
   }
 
-  return collectDeclarations(declaration)[type].standalone === true;
+  // Handle Angular 19+ default standalone behavior for components
+  const declarations = collectDeclarations(declaration);
+  if (Number(VERSION.major) >= 19 && type === 'Component' && declarations[type].standalone === undefined) {
```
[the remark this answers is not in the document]
re-alam: Thank you for noticing it 👍. I updated `isStandalone` to support default standalone for  Directive and Pipe as well.

jensweigele:
I think you need to adapt the unit-tests to get green-light from CI (and approval from @satanTime )

satanTime (approved):

satanTime:
Hi @re-alam, thanks for the fix. 

Could stash all commits into one?
Also, please change the message of the commit to:

fix(a19): respecting default standalone flag from angular 19

Thank you in advance.

satanTime:
Hi @re-alam, could you sign your commit?

the easiest way is SSH signing: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

re-alam:
@satanTime Thank you for approving the change. My commit is now signed. Please feel free to merge when you're able.

libs/ng-mocks/src/lib/common/func.is-standalone.spec.ts
satanTime, line 1: could you move this file to all tests under [tests](https://github.com/help-me-mom/ng-mocks/tree/master/tests)?

re-alam: @satanTime I moved the file to the root of the `tests` folder. Let me know if you prefer a different file/folder name.

libs/ng-mocks/src/index.ts
```diff
@@ -4,6 +4,7 @@ import './lib/common/ng-mocks-stack';
 import './lib/common/ng-mocks-global-overrides';
 
 export * from './lib/common/core.tokens';
+export * from './lib/common/func.is-standalone';
```
satanTime, line 7: It's an internal function and it shouldn't be a part of public interface. 

I think the test should be done in a different way. There are some examples in tests folder which verify different behaviors of different angular versions.

satanTime (changes requested):

libs/ng-mocks/src/index.ts
re-alam: @satanTime Thank you for your feedback. I've looked into the [tests](https://github.com/help-me-mom/ng-mocks/tree/master/tests) folder but couldn't find any tests that specifically cover the feature or logic related to Angular version checks. I only found some tests that skip execution based on the Angular version, like this:

```
if (Number.parseInt(VERSION.major, 10) < 14) {
  it('needs >=a14', () => {
    expect(true).toBeTruthy();
  });
  return;
}
```

However, I need to ensure code coverage for the Angular 19 version check within the `isStandalone` function.

I attempted to mock the `VERSION` object using `spyOnProperty`, but this wasn't possible as the `VERSION` object from `@angular/core` isn't configurable.

Given that `isStandalone` is an internal function, what do you think about moving the test file to the [common](https://github.com/help-me-mom/ng-mocks/tree/master/libs/ng-mocks/src/lib/common) folder? Please let me know if there are any other approaches I should consider.

ct-serensia:
Hi @satanTime & @re-alam,

First, thank you both for your work.

I wanted to let you know that I did a build based from this PR, and it does not seem to be enough to solve all the issue caused by components being standalone by default in Angular 19.

It works for components directly inside the application, but not for non-standalone components from libraries:

```
Error: Unexpected "MockOfMatSelectSearchComponent" found in the "declarations" array of the "MockOfNgxMatSelectSearchModule" NgModule, "MockOfMatSelectSearchComponent" is marked as standalone and can't be declared in any NgModule - did you intend to import it instead (by adding it to the "imports" array)?
```

I see 2 possible problems:
1. they are are always detected as standalone
2. and/or there is a place somewhere in the lib where the `declarations` vs `imports` is not managed correctly

I tried to do a fix myself but without success for now.

For info, when a library is built, there is not anymore a `standalone` property directly inside the component. Instead, there is a `isStandalone` property in what seems to be metadata. For example:

```ts
class MatButton extends MatButtonBase {
    static ɵfac = i0.ɵɵngDeclareFactory({ minVersion: "12.0.0", version: "19.0.0", ngImport: i0, type: MatButton, deps: null, target: i0.ɵɵFactoryTarget.Component });
    static ɵcmp = i0.ɵɵngDeclareComponent({ minVersion: "14.0.0", version: "19.0.0", type: MatButton, isStandalone: true });
}
```

florestankorp:
Shouldn't there be an addition to the 'e2e' folder?

image

GipHub123:
@endlacer Some people don’t seem to understand that most libraries are maintained on a pro bono basis, without any financial compensation 😠   These developers generously dedicate their free time to create and maintain fantastic development tools that benefit us all.

> this has to be merged ASAP. having a testing library block the upgrade of the app to angular 19 is highly unacceptable!

florestankorp:
> this has to be merged ASAP. having a testing library block the upgrade of the app to angular 19 is highly unacceptable!

I understand your frustration, but this doesn't have to be blocking: take this fix, build it and temporarily include it in your bundle to import from there. That's what I'm doing for now and it works. DM me if you have questions or need any help with this.

TomLBarden:
> > this has to be merged ASAP. having a testing library block the upgrade of the app to angular 19 is highly unacceptable!
> 
> @endlacer I understand your frustration, but this doesn't have to be blocking: take this fix, build it and temporarily include it in your bundle to import from there. That's what I'm doing for now and it works. DM me if you have questions or need any help with this.

A bit of guidance on this would be hugely appreciated!

endlacer:
> > this has to be merged ASAP. having a testing library block the upgrade of the app to angular 19 is highly unacceptable!
> 
> @endlacer I understand your frustration, but this doesn't have to be blocking: take this fix, build it and temporarily include it in your bundle to import from there. That's what I'm doing for now and it works. DM me if you have questions or need any help with this.

sorry! did not mean to be so harsh, but we got 4 applications and 2 libraries that all are now blocked bc of something that
1. seems minor
2. already a PR exists to fix it

florestankorp:
> > @endlacer I understand your frustration, but this doesn't have to be blocking: take this fix, build it and temporarily include it in your bundle to import from there. That's what I'm doing for now and it works. DM me if you have questions or need any help with this.
> 
> A bit of guidance on this would be hugely appreciated!
@TomLBarden here you go:

1. Clone 'ng-mocks'
2. Install the dependencies with `npm install`
3. Apply the [changes suggested in this PR](https://github.com/help-me-mom/ng-mocks/pull/10583/files)
4. Build the lib by running: `npm run build:all`
5. Copy the folder `ng-mocks` from `dist` into your project
6. Update the imports in your project to point at the newly imported version of 'ng-mocks'
7. Optional: Disable ESLint for `ng-mocks` by [ignoring the folder](https://eslint.org/docs/latest/use/configure/ignore):
```javascript
// eslint.config.js
export default [
    {
        ignores: ["**/ng-mocks/**"]
    }
];
```

Your tests should be running now with Angular v19 😎

TomLBarden:
@florestankorp Thank you!

> I wanted to let you know that I did a build based on this PR, and it does not seem to be enough to solve all the issue caused by components being standalone by default in Angular 19.
> 
> It works for components directly inside the application, but not for non-standalone components from libraries. I get errors like this one:

@ct-serensia I'm seeing this issue too, did you happen to find a solution for it? I'll continue to investigate this myself, too.

ct-serensia:
@TomLBarden Unfortunately no, I tried some fixes but none work.

My workaround for now is to install back Angular 18 to run unit tests, while everything else (serve, build...) uses Angular 19.

endlacer:
> My workaround for now is to install back Angular 18 to run unit tests, while everything else (serve, build...) uses Angular 19.

this might work local, but not in a CI pipeline

ct-serensia:
@endlacer it works in CI by including the installation step in the CI task.

If it can help, before the unit test command:
```bash
npm install @angular/animations@^18.0.0 @angular/common@^18.0.0 @angular/compiler@^18.0.0 @angular/core@^18.0.0 @angular/forms@^18.0.0 @angular/platform-browser@^18.0.0 @angular/router@^18.0.0 --force && npm install @angular-devkit/build-angular@^18.0.0 @angular/cli@^18.0.0 @angular/compiler-cli@^18.0.0 @angular/platform-browser-dynamic@^18.0.0 typescript@~5.5.0 -D --force
```

You may need to add other libraries which requires Angular 19 specifically (like `@angular/material`).

After the unit test command:
```bash
git restore package.json && git restore package-lock.json && npm ci
```

Of course it is far from ideal, but it is temporary workaround.

TomLBarden:
> My workaround for now is to install back Angular 18 to run unit tests, while everything else (serve, build...) uses Angular 19.

@ct-serensia That's a shame, thanks for the reply though. Looking like I might have to delay this upgrade for now. Unfortunately for us we can't accept the risk of testing against a different version of Angular than we'd be shipping with.

endlacer:
[ct-serensia](https://github.com/ct-serensia) i dont even this this would work. the angular 19 migration removes standalone: true flags from the component. when you then run tests with angular 18, the tests would think that this component is standalone-false, which it is not

szigetib95:
I don't see any update on this for last 2 weeks as it's being worked on, it would be very helpful to get this problem resolved, as it is stopping the complete Nx migration for all our repos.
Thanks!

florianjung79:
For the time being, we continue to declare the standalone components with `standalone: true`. The tests then run as before and as soon as ng-mocks offers a corresponding fix, the global removal of the declaration is a matter of search and replace.

szigetib95:
> For the time being, we continue to declare the standalone components with `standalone: true`. The tests then run as before and as soon as ng-mocks offers a corresponding fix, the global removal of the declaration is a matter of search and replace.

'The global removal of the declaration is a matter of search and replace' - In case your components are standalone, but I need the flag, since its `standalone: false`.

florestankorp:
> standalone:

@florianjung79 Do you mean you set your `MockRender` / `MockBuilder` to `standalone: true`? Or your actual components? Can you provide a code snippet?

florianjung79:
> @florianjung79 Do you mean you set your `MockRender` / `MockBuilder` to `standalone: true`? Or your actual components? Can you provide a code snippet?

The actual components.

szigetib95:
Is there any update on this? Do we need to wait for it at all?

Nvveen:
We really, desperately needs this fix to be merged, because it's blocking us from upgrading to Angular 19. This doesn't seem like a very big fix and solutions already exist, so please merge this ASAP.

ammaraldirawi:
any news on this? is there something we could do to help?

StavNoyAkur8:
In a separate "issue" the owner declared a while back they'll be away until January.
Just today, they said they'll get around to reviewing and updating next week. https://github.com/help-me-mom/ng-mocks/issues/10487#issuecomment-2574735664

We'll have to be patient a little longer. Or you could always fork.

ytchenak:
@re-alam:
The fix isn't sufficient to support non-standalone components in third-party libraries like ngx-translate.  
I applied the following fix to ensure support:
https://github.com/ytchenak/ng-mocks/commit/6fb8cdb9e87b14c044230051a622131f484e8697#diff-0420360326ecb9425d6af7835d8a640bc3b6c77b79137a3a7576b940209cdd86R50

satanTime:
> @re-alam: The fix isn't sufficient to support non-standalone components in third-party libraries like ngx-translate. I applied the following fix to ensure support: [libs/ng-mocks/src/lib/mock/decorate-declaration.ts](https://github.com/ytchenak/ng-mocks/commit/6fb8cdb9e87b14c044230051a622131f484e8697#diff-0420360326ecb9425d6af7835d8a640bc3b6c77b79137a3a7576b940209cdd86R46-R76)
> 
> ```diff
> @@ -46,6 +46,8 @@ export default <T extends Component & Directive>(
>     },
>   params: T,
> ): Component & Directive => {
> 
> +  const standalone = ((source as any).ɵcmp || (source as any).ɵdir || (source as any).ɵpipe)?.standalone;
>   const hasResolver = ngMocksUniverse.config.has('mockNgDefResolver');
>   if (!hasResolver) {
>     ngMocksUniverse.config.set('mockNgDefResolver', new CoreDefStack());
> @@ -61,10 +63,14 @@ export default <T extends Component & Directive>(
>   if (meta.selector !== undefined) {
>     options.selector = meta.selector;
>   }
> 
>   if (meta.standalone !== undefined) {
>     options.standalone = meta.standalone;
> +  } else if (standalone !== undefined) {
> +    options.standalone = standalone;
>   }
> ```

Hi @ytchenak, could you create an example with the failure or explain the issue better?

JLa2r:
This change resolved the same issue I was facing. Thanks to everyone who contributed!

satanTime:
[v14.13.2](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.13.2) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #10751, opened 2025-01-12
fix(a19): correct detection of standalone flag #10306 #10583

### LINKED ISSUE #10306: Bug: Ng19=next - Forgot to flush TestBed? MockRender cannot be used without a reset
### Description of the bug

After upgrading to Angular 19.0.0-next.11 from Angular 18.x my tests my tests that use **MockRender** started getting the error: "Forgot to flush TestBed? MockRender cannot be used without a reset..."  These tests all worked properly before.  I also tried taking the advice and calling **MockRender** like this `fixture = MockRender('', {}, { reset: true });`, but it did not help.

### An example of the bug

You can see an example of the Angular 18 versions here: https://github.com/DDtMM/angular-signal-generators/blob/main/projects/signal-generators/src/lib/signals/tween-signal.spec.ts.  The stack trace pointed to the `MockRender()` calls in the beforeEach.  I'm assuming Angular changed something that affects **ngMocks** since I'm using it in a very minimal fashion for my project.

```
 beforeEach(() => {
    fixture = MockRender(); // <-- error here.
    injector = fixture.componentRef.injector;
  });
```

Don't throw that error.

satanTime:
fixes #10306
closes #10583


satanTime:
[v14.13.2](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.13.2) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #10825, opened 2025-01-18
fix: 9666 Performance

### LINKED ISSUE #9666: Bug: Performance issue
### Description of the bug

`ngMocks.flushTestBed()` function prevents the cleanup-code in angular's TestBed.tearDownTestingModule function. As a result, unit test execution is much slower, because of duplicated, not cleaned up shared host styles (and maybe other objects). (3000 tests, 3-4 minutes instead of 1-1,5 minutes)

### An example of the bug

`ngMocks.flushTestBed()` called from https://github.com/help-me-mom/ng-mocks/blob/master/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts#L56 sets the `_testModuleRef` to `null`
(https://github.com/help-me-mom/ng-mocks/blob/master/libs/ng-mocks/src/lib/mock-helper/mock-helper.flush-test-bed.ts#L7)

This prevents the cleanup of the test module.
https://github.com/angular/angular/blob/main/packages/core/testing/src/test_bed.ts#L735
```
  // If the module ref has already been destroyed, we won't be able to get a test renderer.
    if (this._testModuleRef === null) {
      return;
    }
    // Resolve the renderer ahead of time, because we want to remove the root elements as the very
    // last step, but the injector will be destroyed as a part of the module ref destruction.
    const testRenderer = this.inject(TestComponentRenderer);
    try {
      this._testModuleRef.destroy();
    } catch (e) {
      if (this.shouldRethrowTeardownErrors()) {
        throw e;
      } else {
        console.error('Error during cleanup of a testing module', {
          component: this._testModuleRef.instance,
          stacktrace: e,
        });
      }
    } finally {
      testRenderer.removeAllRootElements?.();
    }
```

Link:

Workaround tests.ts:
Add
```
afterEach(() => { getTestBed().inject(ɵSharedStylesHost).ngOnDestroy(); });
```
https://github.com/angular/angular/issues/31834#issuecomment-732153736

Possible fix in ng-mocks (mock-helper.flush-test-bed.ts)
```
export default (): void => {
  const testBed: any = getTestBed();
  testBed.tearDownTestingModule(); // run cleanup
  testBed._instantiated = false;
  testBed._moduleFactory = undefined;
  testBed._testModuleRef = null;
};
```

lneumeier-visma:
call tearDownTestingModule before setting refs to null

Fixes #9666

### DIFF
--- a/libs/ng-mocks/src/lib/mock-helper/mock-helper.flush-test-bed.ts
+++ b/libs/ng-mocks/src/lib/mock-helper/mock-helper.flush-test-bed.ts
@@ -2,6 +2,9 @@ import { getTestBed } from '@angular/core/testing';
 
 export default (): void => {
   const testBed: any = getTestBed();
+  if (testBed.shouldTearDownTestingModule !== undefined && testBed.shouldTearDownTestingModule()) {
+    testBed.tearDownTestingModule();
+  }
   testBed._instantiated = false;
   testBed._moduleFactory = undefined;
   testBed._testModuleRef = null;
--- a/tests/mock-helper-flush-test-bed/test.spec.ts
+++ b/tests/mock-helper-flush-test-bed/test.spec.ts
@@ -1,0 +1,58 @@
+import { getTestBed } from '@angular/core/testing';
+
+import { ngMocks } from 'ng-mocks';
+
+describe('mock-helper-flush-test-bed', () => {
+  let originalShouldTearDownTestingModule: any;
+  let originalTearDownTestingModule: any;
+
+  beforeAll(() => {
+    const testBed: any = getTestBed();
+    originalShouldTearDownTestingModule =
+      testBed.shouldTearDownTestingModule;
+    originalTearDownTestingModule = testBed.tearDownTestingModule;
+  });
+
+  afterAll(() => {
+    const testBed: any = getTestBed();
+    testBed.shouldTearDownTestingModule =
+      originalShouldTearDownTestingModule;
+    testBed.tearDownTestingModule = originalTearDownTestingModule;
+  });
+
+  it('should execute tearDownTestingModule', () => {
+    const testBed: any = getTestBed();
+
+    testBed.shouldTearDownTestingModule =
+      typeof jest === 'undefined'
+        ? jasmine.createSpy('spyShouldTearDown').and.returnValue(true)
+        : jest.fn().mockReturnValue(true);
+
+    testBed.tearDownTestingModule =
+      typeof jest === 'undefined'
+        ? jasmine.createSpy('spyTearDown')
+        : jest.fn();
+
+    ngMocks.flushTestBed();
+    expect(testBed.tearDownTestingModule).toHaveBeenCalledTimes(1);
+  });
+
+  it('should not execute tearDownTestingModule', () => {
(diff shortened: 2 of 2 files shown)


libs/ng-mocks/src/lib/mock-helper/mock-helper.flush-test-bed.ts
```diff
@@ -2,6 +2,7 @@ import { getTestBed } from '@angular/core/testing';
 
 export default (): void => {
   const testBed: any = getTestBed();
+  testBed.tearDownTestingModule();
```
satanTime, line 5: Could you add logic to call it only if it exists?

satanTime (changes requested):

libs/ng-mocks/src/lib/mock-helper/mock-helper.flush-test-bed.ts
lneumeier-visma: I added the conditional logic and added some tests to get 100% code coverage, unfortunately they are not running on circle ci. I looks like `jasmine` is not available.
Do you have an idea to get the tests running?

lneumeier: Okay, I found the solution :-)

libs/ng-mocks/src/lib/mock-helper/mock-helper.flush-test-bed.ts
```diff
@@ -2,6 +2,9 @@ import { getTestBed } from '@angular/core/testing';
 
 export default (): void => {
   const testBed: any = getTestBed();
+  if (testBed.shouldTearDownTestingModule != null && testBed.shouldTearDownTestingModule()) {
```
satanTime, line 5: Interesting why lint hasn't throw an error here. 

It should be !== undefined

satanTime (changes requested):

satanTime:
Looks good. Thanks. 

Can you combine commits into one and name it like:

fix(core): <deacription> #<issue number>

???

lneumeier-visma:
> Looks good. Thanks.
> 
> Can you combine commits into one and name it like:
> 
> fix(core): #
> 
> ???

Done.

satanTime (approved):

satanTime:
Thank you, sir!

satanTime:
Hi @lneumeier-visma, could you sign commits?

you can do that with your ssh key: https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account

lneumeier-visma:
> Hi @lneumeier-visma, could you sign commits?
> 
> you can do that with your ssh key: https://docs.github.com/en/authentication/connecting-to-github-with-ssh/adding-a-new-ssh-key-to-your-github-account

Yes, thank you for your patience.

satanTime:
Thank you for the contribution!!!

satanTime:
[v14.13.3](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.13.3) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

Issue #10752, opened 2025-01-12
Add ngx-translate coverage in e2e


satanTime:
The fix isn't sufficient to support non-standalone components in third-party libraries like ngx-translate.  
I applied the following fix to ensure support:
[libs/ng-mocks/src/lib/mock/decorate-declaration.ts](https://github.com/ytchenak/ng-mocks/commit/6fb8cdb9e87b14c044230051a622131f484e8697#diff-0420360326ecb9425d6af7835d8a640bc3b6c77b79137a3a7576b940209cdd86R46-R76)
```diff
@@ -46,6 +46,8 @@ export default <T extends Component & Directive>(
    },
  params: T,
): Component & Directive => {

+  const standalone = ((source as any).ɵcmp || (source as any).ɵdir || (source as any).ɵpipe)?.standalone;
  const hasResolver = ngMocksUniverse.config.has('mockNgDefResolver');
  if (!hasResolver) {
    ngMocksUniverse.config.set('mockNgDefResolver', new CoreDefStack());
@@ -61,10 +63,14 @@ export default <T extends Component & Directive>(
  if (meta.selector !== undefined) {
    options.selector = meta.selector;
  }

  if (meta.standalone !== undefined) {
    options.standalone = meta.standalone;
+  } else if (standalone !== undefined) {
+    options.standalone = standalone;
  }

```

_Originally posted by @ytchenak in https://github.com/help-me-mom/ng-mocks/issues/10583#issuecomment-2577323646_

KernelFolla:
Hi, I'm trying to fix it too. 

Have you noticed that pipes are now standalone by default?

I don't understand a thing - why not use the isStandalone method from @angular/core to check this? It would seem to be the most direct way to determine a component|directive|pipe standalone status.

satanTime:
I'll backport its implementation. The issue here is that this function doesn't exist in older angular versions and will break ng-mocks at project compilations.

KernelFolla:
I've made this, 
https://github.com/help-me-mom/ng-mocks/pull/10819

It's late and I'm not sure what I did - there's definitely an ugly part in there, but it seems to work for me. Maybe it can help you

ytchenak:
@satanTime,
please see https://stackblitz.com/~/github.com/ytchenak/t1-ng-mocks  as example.
I’m using **@ngx-translate/core@15.0.0**, which uses the non-standalone version to reproduce the issue.,

```
nx test t1-ng-mocks

[...]

Issue #4032, opened 2022-10-29
Mocking a Standalone Component without a Selector does not mock the imports


Simon-TheHelpfulCat:
A minimal reproduction:

```typescript
import { CommonModule } from "@angular/common";
import { Component } from "@angular/core";
import { RouterModule } from "@angular/router";
import { MockBuilder, MockRender } from "ng-mocks";

@Component({
  selector: "my-test-one",
  standalone: true,
  imports: [CommonModule, RouterModule],
  template: `<a [routerLink]="['test']">Test</a>`
})
class TestOneComponent {}

@Component({
  standalone: true,
  imports: [CommoBoth ytestsnModule, RouterModule],
  template: `<a [routerLink]="['test']">Test</a>`
})
class TestTwoComponent {}

describe("with selector", () => {
  beforeEach(() => MockBuilder(TestOneComponent));

  it("should render", () => {
    expect(MockRender(TestOneComponent).point.componentInstance).toBeDefined();
  });
});

describe("witout selector", () => {
  beforeEach(() => MockBuilder(TestTwoComponent));

  it("should render", () => {
    expect(MockRender(TestTwoComponent).point.componentInstance).toBeDefined();
  });
});
```
Both tests pass, but looking in the console you see:

> NG0303: Can't bind to 'routerLink' since it isn't a known property of 'a' (used in the 'MockMiddlewareTestTwoComponent' component template).

Adding a selector to TextTwoComponent or adding ```.mock(RouterModule))``` to the ```MockBuilder``` removes this message.

This minimal reproduction is reproduced [in Code Sandbox](https://codesandbox.io/s/objective-sun-cyh5nx?file=/src/test.spec.ts).

satanTime:
Hi @Simon-TheHelpfulCat,

thanks for the report!

satanTime:
[v14.3.2](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.3.2) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #10826, opened 2025-01-18
feat: standalone pipes and legacy external modules support in angular 19

KernelFolla:
this PR solves two things:

1. angular standalone components in external legacy modules ( https://github.com/help-me-mom/ng-mocks/issues/10752 )
2. standalone pipes 

I've tested the following code in my project without issues but I'm not confident on how to run and fix tests with karma and e2e

I hope this PR somehow contributes towards the definitive solution for supporting Angular 19🙏

### DIFF
--- a/libs/ng-mocks/src/lib/common/decorate.mock.ts
+++ b/libs/ng-mocks/src/lib/common/decorate.mock.ts
@@ -18,4 +18,9 @@ export default (mock: AnyType<any>, source: AnyType<any>, configInput: ngMocksMo
       }
     : configInput;
   coreDefineProperty(mock.prototype, '__ngMocksConfig', config);
+  if ((mock as any)?.ɵmod?.declarations) {
+    (mock as any).ɵmod.declarations
+      .filter((x: any) => x?.decorators?.[0]?.args?.[0] && x?.decorators?.[0]?.args?.[0].standalone === undefined)
+      .map((x: any) => (x.decorators[0].args[0].standalone = false));
+  }
 };
--- a/libs/ng-mocks/src/lib/mock-pipe/mock-pipe.ts
+++ b/libs/ng-mocks/src/lib/mock-pipe/mock-pipe.ts
@@ -9,6 +9,7 @@ import funcImportExists from '../common/func.import-exists';
 import { isMockNgDef } from '../common/func.is-mock-ng-def';
 import { Mock } from '../common/mock';
 import ngMocksUniverse from '../common/ng-mocks-universe';
+import decorateDeclaration from '../mock/decorate-declaration';
 import returnCachedMock from '../mock/return-cached-mock';
 import helperMockService from '../mock-service/helper.mock-service';
 
@@ -36,7 +37,9 @@ const getMockClass = (pipe: Type<any>, transformValue?: PipeTransform['transform
   const config = ngMocksUniverse.config.get(pipe);
   const transform = transformValue ?? config?.defValue?.transform;
   const mock = extendClass(Mock);
-  Pipe(coreReflectPipeResolve(pipe))(mock);
+  const meta = coreReflectPipeResolve(pipe);
+
+  Pipe(decorateDeclaration(pipe, mock, meta, {}) as Pipe)(mock);
   decorateMock(mock, pipe, {
     init: (instance: PipeTransform) => {
       if (transform) {
--- a/libs/ng-mocks/src/lib/mock-render/func.reflect-template.ts
+++ b/libs/ng-mocks/src/lib/mock-render/func.reflect-template.ts
@@ -30,7 +30,7 @@ const registerTemplateMiddleware = (template: AnyType<any>, meta: Directive): vo
     // nothing to do
   }
 
-  const standalone = (meta as any).__ngMocksStandalone === true;
+  const standalone = (meta as any).standalone === true;
   (isNgDef(template, 'c') ? Component : Directive)({
     ...meta,
     ...set,
--- a/libs/ng-mocks/src/lib/common/core.reflect.directive-resolve.ts
+++ b/libs/ng-mocks/src/lib/common/core.reflect.directive-resolve.ts
@@ -1,4 +1,4 @@
-import { Component, Directive, NgModule } from '@angular/core';
+import { Component, Directive, NgModule, Pipe } from '@angular/core';
 
 import collectDeclarations from '../resolve/collect-declarations';
 
@@ -8,6 +8,7 @@ export default (
   def: any,
 ): Directive &
   Partial<Component> &
+  Pipe &
   NgModule & {
     hostBindings?: Array<[string, any]>;
     hostListeners?: Array<[string, any, any]>;
@@ -21,6 +22,9 @@ export default (
     }
     if (declaration.Directive) {
       return declaration.Directive;
+    }
+    if (declaration.Pipe) {
+      return declaration.Pipe;
     }
 
     throw new Error('Cannot resolve declarations');
--- a/libs/ng-mocks/src/lib/common/func.is-standalone.ts
+++ b/libs/ng-mocks/src/lib/common/func.is-standalone.ts
@@ -1,6 +1,3 @@
-import ngMocksUniverse from '../common/ng-mocks-universe';
-import collectDeclarations from '../resolve/collect-declarations';
-
 import { getNgType } from './func.get-ng-type';
 
 /**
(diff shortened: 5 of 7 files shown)


Martinspire:
Tests seem to fail, any idea on that?

vzakharov-rxnt:
This 100% fixed all the relevant issues in my project. It's an enterprise class frontend app with probably 500k LOC of frontend code, legacy AJS + A19 hybrid. We should merge your PR ASAP.

kheos31:
Hello! Do you find a way to correct the failing tests @KernelFolla ? Thanks for your contribution anyway, hoping it will be merged ASAP regarding of issues linked.

KernelFolla:
> Hello, do you find a way to correct the failing tests @KernelFolla ? Thanks for your contribution anyway, hoping it will be merged ASAP regarding of issues linked.

can I say I don't know how to do it? :) anyway I'll try this weekend, any suggestions and help are welcome.

sumegha26:
Any plans to merge this helpful and much needed change ?

kheos31:
> > Hello, do you find a way to correct the failing tests @KernelFolla ? Thanks for your contribution anyway, hoping it will be merged ASAP regarding of issues linked.
> 
> can I say I don't know how to do it? :) anyway I'll try this weekend, any suggestions and help are welcome.

Yes you can! 🤣 I'm not used to work with NgMocks source code and I think @satanTime or maybe an usual contributor could help on it. As I read in your changed code and in the original code, I think there is a missing concept in isStandalone function. 

There is no way currently for this function to know the using context of an entity (component, directive, pipe). Without this information, how NgMocks could know if an entity is standalone or not? If this function could know if the dependency tree entity to mock is declared by a Module, we could deduce that is not a standalone component. It could be summed up as : 

```javascript
if (isEntityDeclaredInModule) {    // <--- treatment for component/directive/pipe declared into module, concept currently absent
    return false;
} else {
   // do your current modified isStandalone treatment
}
```

Maybe i'm wrong but as NgMocks knows the full entity dependency tree, I think you could find this information and add it to the isStandalone function (or call the isStandalone function just in "else" case).

Good luck @KernelFolla 😉

joepvl:
In general, this PR looks correct to me. So far the CI seems to have failed on a technical issue, not an actual test failure. Can you try re-running the build like the CircleCI UI suggests? Screenshot:
image

FrankMerema:
> There is no way currently for this function to know the using context of an entity (component, directive, pipe). Without this information, how NgMocks could know if an entity is standalone or not? If this function could know if the dependency tree entity to mock is declared by a Module, we could deduce that is not a standalone component. It could be summed up as :
> 
> ```js
> if (isEntityDeclaredInModule) {    // <--- treatment for component/directive/pipe declared into module, concept currently absent
>     return false;
> } else {
>    // do your current modified isStandalone treatment
> }
> ```
> 
> Maybe i'm wrong but as NgMocks knows the full entity dependency tree, I think you could find this information and add it to the isStandalone function (or call the isStandalone function just in "else" case).

@kheos31 To be honest I don't fully understand what you're referring to here? Could you explain it a bit more?

I'm currently looking at the tests and I found something that is probably not correct in this PR. I'm trying to understand how everything works together in order to fix the code and with that the code :)

libs/ng-mocks/src/lib/mock-pipe/mock-pipe.ts
```diff
@@ -36,7 +37,9 @@ const getMockClass = (pipe: Type<any>, transformValue?: PipeTransform['transform
   const config = ngMocksUniverse.config.get(pipe);
   const transform = transformValue ?? config?.defValue?.transform;
   const mock = extendClass(Mock);
-  Pipe(coreReflectPipeResolve(pipe))(mock);
+  const meta = coreReflectPipeResolve(pipe);
+
+  Pipe(decorateDeclaration(pipe, mock, meta, {}) as Pipe)(mock);
```
FrankMerema, line 42: Because of the decorateDeclaration call here, the decorateMock will be called twice. Once inside the method and once on the line below (43). Therefore the contents of the mock object are overwritten the second time it's called and breaking the tests.

libs/ng-mocks/src/lib/common/decorate.mock.ts
```diff
@@ -18,4 +18,9 @@ export default (mock: AnyType<any>, source: AnyType<any>, configInput: ngMocksMo
       }
     : configInput;
   coreDefineProperty(mock.prototype, '__ngMocksConfig', config);
+  if ((mock as any)?.ɵmod?.declarations) {
```
FrankMerema, line 21: Hi @KernelFolla, could you maybe elaborate a bit on this piece of code? I don't fully understand the purpose of it?

libs/ng-mocks/src/lib/mock-render/func.reflect-template.ts
```diff
@@ -30,7 +30,7 @@ const registerTemplateMiddleware = (template: AnyType<any>, meta: Directive): vo
     // nothing to do
   }
 
-  const standalone = (meta as any).__ngMocksStandalone === true;
+  const standalone = (meta as any).standalone === true;
```
FrankMerema, line 33: Why is this part needed? It breaks the test for https://github.com/help-me-mom/ng-mocks/issues/4032. Locally I've reverted this line and I don't see any difference.

kheos31:
> > There is no way currently for this function to know the using context of an entity (component, directive, pipe). Without this information, how NgMocks could know if an entity is standalone or not? If this function could know if the dependency tree entity to mock is declared by a Module, we could deduce that is not a standalone component. It could be summed up as :
> > ```js
> > if (isEntityDeclaredInModule) {    // <--- treatment for component/directive/pipe declared into module, concept currently absent
> >     return false;
> > } else {
> >    // do your current modified isStandalone treatment
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Maybe i'm wrong but as NgMocks knows the full entity dependency tree, I think you could find this information and add it to the isStandalone function (or call the isStandalone function just in "else" case).
> 
> @kheos31 To be honest I don't fully understand what you're referring to here? Could you explain it a bit more?
> 
> I'm currently looking at the tests and I found something that is probably not correct in this PR. I'm trying to understand how everything works together in order to fix the code and with that the code :)

Hello, I just said that NgMocks have to treat with the same maneer standalone components with "standalone: true" (old way) and without (Angular 19 way). NgMocks also have to treat the non standalone components differently. 
The question is: how to differentiate these cases? Currently, as I could saw in the source code, I think NgMocks just cant. The only way to achieve the "isStandalone" check, IMO, is by knowing if a component is declared by a Module which NgMocks tried to mock before he tried to mock the component.

```
@NgModule({
    declarations: [
        NotStandaloneComponent,
    ],
    exports: [
        NotStandaloneComponent,
    ],
})
export class MyModule {}
```

It's just impossible to use a NotStandaloneComponent in another SComponent without this SComponent imports MyModule containing the NotStandaloneComponent. Dependency tree:

`SComponent <--imports-- MyModule <--declares-- NotStandaloneComponent` 

Thus, using NgMocks, you'll necessarily call MockModule(MyModule) in the unit test of SComponent. Doing that giving us our entry point to the tree dependency which have to be mocked, and we therefore have the information we need to deduct our "isStandalone" check.

Martinspire:
What if there was a parameter to make sure ngmocks knows something will not be a standalone component? We all probably know the component it is failing on and having `MockModule(SomeModule, <some config>)` or `MockComponent(SomeComponent, <some config>)` is fine by me. As long as I'm at least able to do such a thing. Its still much better than having to do a custom component just because ngmocks can't figure it out.

FrankMerema:
In the [ng-mocks-universe.ts](https://github.com/help-me-mom/ng-mocks/blob/04f89246f257b63435a7a4bb4bcb53f7e2a580ec/libs/ng-mocks/src/lib/common/ng-mocks-universe.ts#L65C10-L65C19) there is a boolean value that checks if you're in the world of `defaultStandalone` AKA >= NG19 . We might be able to use this value to do the correct if/else right? In that case you don't have to do difficult checks :)

kheos31:
> In the [ng-mocks-universe.ts](https://github.com/help-me-mom/ng-mocks/blob/04f89246f257b63435a7a4bb4bcb53f7e2a580ec/libs/ng-mocks/src/lib/common/ng-mocks-universe.ts#L65C10-L65C19) there is a boolean value that checks if you're in the world of `defaultStandalone` AKA >= NG19 . We might be able to use this value to do the correct if/else right? In that case you don't have to do difficult checks :)
> 
> It will look like this if I'm correct: image

Unfortunately no, because external libraries could be "Angular 19" compatible but not migrated as "default standalone". It's the problem here, and why I suggested previous idea to manage this check.

libs/ng-mocks/src/lib/common/decorate.mock.ts
KernelFolla: when mocking angular modules, this code ensures that declarations which don't explicitly specify **standalone: true** are marked as **standalone: false**

libs/ng-mocks/src/lib/mock-pipe/mock-pipe.ts
KernelFolla: I noticed that line 42 (the line starting with 'Pipe') appears to be redundant and could potentially be removed without consequences. I plan to verify this hypothesis tomorrow when I'll have access to the code to test this change.

anyway changes on this file are all made in order to support standalone pipes

libs/ng-mocks/src/lib/mock-render/func.reflect-template.ts
KernelFolla: Honestly I don't remember, tomorrow I'll try this too, reverting on my codebase

Issue #11001, opened 2025-02-12
Bug: Angular 19 setInput with InputSignal on test


handja:
Hi,

I encounter an issue on the angular unit test when I want to set a value to an input.

Inside my component, i've only one input : 

```
export class Component {
  protected items = input<Item[]>([]);
}
```

Inside my test I want to set the items value with he `setInput()` method.
`fixture.componentRef.setInput('items', items);`

But this issue is encountered when I use this method.

**Error: NG0303: Can't set value of the 'items' input on the 'r' component. Make sure that the 'items' property is annotated with @Input() or a mapped @Input('items') exists.**

I'm using :

- ng-mock: 14.13.2
- jasmine: 5.1.4
- Angular: 19

GipHub123:
@handja Could you provide a complete example of how you set up your test cases?

handja:
TS File
```
import { Component, input } from '@angular/core';

@Component({
  selector: 'esac-test',
  template: '',
})
export class TestComponent {
  readonly count = input(0);
}
```

Spec File
```
import { TestComponent } from './test.component';
import { MockBuilder, MockRender, MockedComponentFixture } from 'ng-mocks';

fdescribe('TestComponent', () => {
  let component: TestComponent;
  let fixture: MockedComponentFixture<TestComponent>;

  beforeEach(async () => {
    await MockBuilder(TestComponent);
    fixture = MockRender(TestComponent);
    component = fixture.point.componentInstance;
  });

  it('should create', () => {
    fixture.componentRef.setInput('count', 1);
    expect(component).toBeTruthy();
  });
});

```

StreetCube:
I'm having the same issue, can't set any of the inputs with ComponentRef.setInput() getting the same error message.

jeppetn:
We are having same issue, is there a timeline for this?

cheng93:
workaround has worked for me

```ts
import { SIGNAL } from "@angular/core/primitives/signals";

component.items[SIGNAL].applyValueToInputSignal(component.items[SIGNAL], items);
```

source

https://github.com/angular/angular/blob/main/packages/core/src/render3/apply_value_input_field.ts#L11

satanTime:
[v14.16.0](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.16.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #8818, opened 2024-04-24
feat: support signal input from angular 17.1

andreandersson:
Adds a custom transformer to transform Angular Code, specifically, signals (input, query and models) to behave better in jit-environment.

Solves #7976

PR #11527, opened 2025-05-07
fix(helper): add possibility to set typed input

JosefBredereck:
Add helper which extracts all input properties that are of type `InputSignal` or `ModelSignal` from the component type and allows setting their values in a type-safe manner.

Aligns as a extension to #8818, which allows setting the default inputs.
This might not work with older angular versions that do not support `InputSignal`s.
How would you handle this case @satanTime when something new is introduced? Should it be a breaking change?

depends on #8818 or whatever the solution for #11001 might be


scriptforfivem:
Great implementation of this feature. I particularly like the approach to error handling.

JosefBredereck:
Trying to add it to angular directly makes more sence I guess.

PR #14004, opened 2026-06-02
fix(MockRender): type signal input bindings #14003

### LINKED ISSUE #14003: Feature Request: sensible signal input handling for MockRender(Component) /  DefaultRenderComponent

When using `MockRender(Component)` the params take input signals as-is, which breaks usage.

Example:

```ts
@Component()
export class MyComponent {
  myInput = input<string>(null);
}

const params = MockRender(MyComponent).componentInstance;
// p has type `{myInput: InputSignal<string>}`
```
The sensible type would be:
```ts
const params = MockRender(MyComponent).componentInstance;
// p has type `{myInput: string | undefined}`
```

### Proposed solution

`DefaultRenderComponent` could be extended like this:

```ts
import type { InputSignalWithTransform } from '@angular/core';  
  
type InputBindingType<T> =  
	T extends InputSignalWithTransform<any, infer WriteT>  
	? WriteT  
	: T;  
  
export type DefaultRenderComponent<MComponent> = {  
	[K in keyof MComponent]: InputBindingType<MComponent[K]>;  
};
```

#### Breaking Change
Of course, technically, this would be a breaking change, but I think this is what every user would expect anyways. Also the old type (probably) did not really work, as assigning an singal instance to a signal input does not make sense.

### Additional context

Relevant docs:
> When MockRender(Component) is used without params then fixture.componentInstance controls @Inputs and @Outputs of the component. That lets it trigger the correct lifecycle hooks.

This then could be updated to
> When MockRender(Component) is used without params then fixture.componentInstance controls @Inputs and @Outputs of the component. That lets it trigger the correct lifecycle hooks. For signal inputs it automatically uses the binding/write type.

satanTime:
Fixes #14003

Why:

- `MockRender(Component)` exposes a generated wrapper through `fixture.componentInstance`, but signal input fields were typed as signal instances instead of the values accepted by Angular bindings.
- ng-mocks supports legacy Angular and TypeScript versions that cannot parse the conditional types needed for signal-aware declarations.
- Karma passed a request-relative tsconfig path to ts-loader, making CI depend on which module initialized the loader first.

What:

- Map optional, required, nullable, and transformed signal inputs to their binding/write types while preserving ordinary fields and non-input signals.
- Keep `index.d.ts` as the legacy default and select `index.modern.d.ts` for TypeScript 3.1 and newer, without exposing internal package imports.
- Generate both declaration bundles from explicit source entrypoints.
- Resolve the root Karma tsconfig before configuring webpack loaders.
- Add the issue regression and document the public `MockRender` behavior.

Where:

- `libs/ng-mocks/src/lib/mock-render`
- `libs/ng-mocks/src/index*.ts`
- `libs/ng-mocks/package.json`
- `karma.conf.ts`
- `tests/issue-14003/test.spec.ts`
- `docs/articles/api/MockRender.md`


satanTime:

PR #14318, opened 2026-07-11
chore(root): remove TypeScript 5 alias

satanTime:
## Summary

- remove the root `typescript5` npm alias
- let `ts-loader` use the root TypeScript 6 compiler
- silence TypeScript 6 deprecation diagnostics while preserving the legacy ES5 bundle

## Why

The second compiler existed only to avoid TypeScript 6 errors for deprecated ES5 build options. TypeScript 6 can still produce the bundle with `ignoreDeprecations: "6.0"`, so keeping a duplicate compiler dependency—and the odd automated updates it creates—is unnecessary.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -73,7 +73,6 @@
         "ts-node": "10.9.2",
         "tsconfig-paths-webpack-plugin": "4.2.0",
         "typescript": "6.0.3",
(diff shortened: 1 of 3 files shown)


satanTime:

PR #14425, opened 2026-07-27
build: guard Dependabot major updates

satanTime:
## Summary

- update the Dependabot preparation skill to detect and close `package.json` SemVer-major updates before approval or auto-merge
- configure all 26 npm update entries to ignore `version-update:semver-major`
- leave Docker update behavior unchanged

## Why

The previous workflow treated clear Dependabot dependency updates as eligible without checking whether a declared dependency crossed a major version. Grouped updates could therefore carry incompatible Angular tooling or npm upgrades into version-specific projects.

## Impact

Scheduled npm version updates will no longer propose SemVer-major upgrades. The preparation workflow remains a second guard: if a major update still reaches review through an edge case, it is closed instead of approved or queued for auto-merge.


satanTime:

PR #14489, opened 2026-08-01
fix(MockRender): respect implicit OnPush strategy #14157

### LINKED ISSUE #14157: Bug: ChangeDetectionStrategy.OnPush being default is not respected with Angular 22
### Description of the bug

With Angular 22, Components that do not explicitly define a ChangeDetectionStrategy use ChangeDetectionStrategy.OnPush.
Somewhere in the library however components without an explicit ChangeDetectionStrategy get set to use ChangeDetectionStrategy.Default or .Eager (the default with Angular <=21), leading to those Components using the wrong change detection strategy for tests

### An example of the bug
Was not able to get the stackblitz to work, but the following test shows the issue:

```typescript
import { ChangeDetectionStrategy, Component, VERSION } from '@angular/core';

import { MockBuilder, MockRender } from 'ng-mocks';

@Component({
  selector: 'target-14157',
  template: '<p></p>',
})
class TargetComponent {

}
describe('issue-14157', () => {
  if (Number.parseInt(VERSION.major, 10) < 22) {
    it('needs a22+', () => {
      expect(true).toBeTruthy();
    });

    return;
  }

  beforeEach(() => MockBuilder(TargetComponent));

  it('respects the default OnPush change detection', () => {
    const fixture: any = MockRender(TargetComponent)
    const annotation =
      fixture.componentInstance?.constructor?.__annotations__?.[0];
    expect(annotation.changeDetection).toEqual(ChangeDetectionStrategy.OnPush)
  });
});

```

Expected: The test should pass on Angular 22, since ChangeDetectionStrategy.OnPush is default
Actual: The test fails because annotation.changeDetection === ChangeDetectionStrategy.Eager

EDITED: Previous report was wrong, absolute brain fart

satanTime:
## Why

Angular 22 defaults components with an omitted `changeDetection` setting to OnPush. Angular's decorator annotations report these components the same way as explicit Eager components, so `MockRender` forced the rendered component's change detector and hid the effective OnPush behavior.

## What

- Read the effective strategy from the compiled component definition before forcing point change detection.
- Add cross-version regression coverage for the implicit framework default, new input references, and explicit Default behavior.
- Mark existing test fixtures that intentionally depend on eager updates as explicit Default components.
- Allow test fixtures to opt into eager change detection without violating the production lint policy.

## Impact

`MockRender` now preserves Angular 22's implicit OnPush semantics while retaining eager/default behavior on Angular 5–21 and for explicitly eager components.

Fixes #14157

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -288,6 +288,8 @@ export default defineConfig([
   {
     files: ['**/*.spec.ts'],
     rules: {
+      // Test fixtures may intentionally exercise eager change detection.
+      '@angular-eslint/prefer-on-push-component-change-detection': 'off',
       'max-lines': 'off',
       'max-lines-per-function': 'off',
       'no-useless-assignment': 'off',
--- a/examples/MockComponent/test.spec.ts
+++ b/examples/MockComponent/test.spec.ts
@@ -1,5 +1,6 @@
 import { CommonModule } from '@angular/common';
 import {
+  ChangeDetectionStrategy,
   Component,
   ContentChild,
   EventEmitter,
@@ -30,6 +31,7 @@ class ChildComponent {
 }
 
 @Component({
+  changeDetection: ChangeDetectionStrategy.Default,
   selector: 'target-mock-component',
   ['standalone' as never /* TODO: remove after upgrade to a14 */]: false,
   template: `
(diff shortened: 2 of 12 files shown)


satanTime:

PR #14491, opened 2026-08-02
fix(MockRender): preserve callable input values #13089

satanTime:
## Why

`MockRender` treated every callable parameter as a method and cloned it to preserve invocation context. Angular Signal Forms field trees are callable proxy values, so cloning them loses their proxy behavior and identity when passed through a component input.

## What

- Derive the wrapper's public input binding names from Angular metadata.
- Pass callable input values through unchanged while retaining method and output-handler rebinding.
- Add a cross-version callable-input regression and an Angular 21+ Signal Forms regression.

## Impact

`MockRender` now preserves function-valued inputs, including Signal Forms `FieldTree` proxies, without framework-specific special cases or changes to method binding behavior.

Fixes #13089


satanTime:

PR #14493, opened 2026-08-02
fix(TestBed): support compiled platform modules #12725

satanTime:
## Why

Angular's legacy Karma builder creates its internal `TestModule` by adding compiled module definitions to a class without decorator metadata. The ng-mocks global platform scan treated every TestBed platform module as an explicitly supplied module and used strict reflection, so it rejected that builder-owned class before `MockProvider` configuration could complete.

## What

- Use optional metadata reflection while scanning TestBed platform modules for global overrides.
- Skip compiled-only internal modules that cannot participate in ng-mocks overrides while retaining strict validation for explicit ng-mocks module inputs.
- Add an Angular 21+ regression covering `MockProvider` with a runtime-compiled `TestModule`.

## Impact

Angular 21 legacy Karma tests can configure mocked providers without the false missing-decorator error. Explicitly passing an undecorated declaration to ng-mocks continues to report the existing validation error.

Fixes #12725


satanTime:

PR #14496, opened 2026-08-02
fix(MockRender): support setInput for signal inputs #11001

### LINKED ISSUE #11001: Bug: Angular 19 setInput with InputSignal on test

satanTime:
## Why

`fixture.componentRef` belongs to the generated `MockRender` wrapper rather than the rendered component. The wrapper stored binding values but did not declare them as Angular inputs, so `ComponentRef.setInput(...)` rejected input names with NG0303. Angular also reflects signal inputs as structured metadata, which caused the generated template to omit their bindings.

## What

- Normalize reflected input metadata to public binding names when generating the wrapper template.
- Declare the generated bindings as ordinary wrapper inputs while preserving aliases and `params` filtering.
- Add cross-version regression coverage for unaliased, aliased, transformed, and params-filtered signal inputs.
- Document how wrapper inputs connect `ComponentRef.setInput(...)` to the rendered `InputSignal`.

## Impact

`fixture.componentRef.setInput(...)` can now update signal inputs through their public binding names. Change detection forwards the wrapper value to the real `InputSignal`, where aliases and transforms retain their Angular behavior and transforms run exactly once. Existing direct wrapper binding access and filtered binding behavior remain unchanged.

Fixes #11001


satanTime:

PR #14490, opened 2026-08-02
fix(MockComponent): preserve signal input behavior #13671

satanTime:
## Why

Mocked components registered signal inputs as ordinary Angular inputs and did not initialize the corresponding `InputSignal` fields. Angular bindings therefore replaced those fields with their bound values, so calling an input such as `name()` failed at runtime.

## What

- Preserve signal, required, alias, and transform metadata while collecting and decorating mock inputs.
- Initialize mocked signal inputs with Angular's `input()` and `input.required()` factories.
- Keep `ngMocks.input(...)` value-oriented by unwrapping inputs identified as signals.
- Add cross-version regression coverage for callable signal inputs and aliased bindings.

## Impact

Signal inputs on mocked components now retain the same callable API as the real component while still receiving values through Angular bindings. Existing tests that inspect bindings through `ngMocks.input(...)` continue to receive the bound value.

Fixes #13671

PR #14499, opened 2026-08-02
test(MockComponent): cover required signal inputs #8887

### LINKED ISSUE #8887: Bug:
I've noticed that Signal Inputs are just treated as Decorator Input. 

> ng-mocks@14.12.2

I have the following component: 
```ts 
@Component({
  selector: 'app-list-item',
  standalone: true,
  templateUrl: './list-item.component.html',
  styleUrl: './list-item.component.scss',
})
export class ListItemComponent { 
  task = input.required<Task>();
}
```

I mock the component using MockComponent function: 

```ts
 await TestBed.configureTestingModule({
      imports: [ListComponent],
      providers: [
        {
          provide: TasksService,
          useValue: tasksServiceStub,
        },
      ],
    })
      .overrideComponent(ListComponent, {
        remove: {
          imports: [ListItemComponent],
        },
        add: {
          imports: [MockComponent(ListItemComponent)], // I mock here
        },
      })
      .compileComponents();
```

And my test try to access the Singal Input from `ListItemComponent` to assert its value: 

```ts
 expect((<ListItemComponent>item.componentInstance).task()).toEqual(completedTasksStub[index]);
```

But the test breaks because the `task` property (it's a Signal Input) is not a function.

The complete error: 
`TypeError: completedItemDebugEl.componentInstance.task is not a function` 

Logging the `<ListItemComponent>item.componentInstance).task` in the test, I see that it's a property

`task: { id: '1', name: 'Buy milk', completed: true }`

Working with Signal Inputs, we need to test it as a Signal and not as a property as it was with Decorator Inputs. 

Any news about that @satanTime? 

Thank for your amazing work btw

satanTime:
Why:

- #8887 reports that `MockComponent` turns an `input.required()` field into a plain property when the mock replaces a standalone import through `TestBed.overrideComponent`.
- The behavior is addressed by #14490, but the existing coverage does not reproduce the required-input override path from the report.

What:

- Add a focused regression that replaces the standalone child with `MockComponent`.
- Assert that the bound required input remains a signal and is still callable on the mocked component instance.

Where:

- Add `tests/issue-8887/test.spec.ts`.
- Gate the spec to Angular targets with signal-input support in `test-spread.conf`.

Closes #8887

### DIFF
--- a/test-spread.conf
+++ b/test-spread.conf
@@ -62,6 +62,7 @@ file|tests/issue-13089/test.spec.ts|features=signalForms
 file|tests/issue-12725/test.spec.ts|versions=21+
 file|tests/issue-13671/test.spec.ts|features=signalInputs
 file|tests/issue-14003/test.spec.ts|features=signalInputs
+file|tests/issue-8887/test.spec.ts|features=signalInputs
 file|tests/issue-6143/*.ts|versions=15+
 file|tests/issue-7938/*.ts|versions=15+
 file|tests/issue-11324/test.spec.ts|versions=15+
--- a/tests/issue-8887/test.spec.ts
+++ b/tests/issue-8887/test.spec.ts
@@ -1,0 +1,77 @@
+import {
+  Component,
+  input,
+  isSignal,
+  reflectComponentType,
+} from '@angular/core';
+import { TestBed } from '@angular/core/testing';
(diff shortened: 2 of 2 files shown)


satanTime:

PR #14500, opened 2026-08-02
test(MockComponent): cover aliased signal inputs #9684

### LINKED ISSUE #9684: Bug: Input Signal with alias seems not to be picked up
### Description of the bug

I've made this (internal) change in one of my components "A" which leaves the public api of the component unchanged.
image

The component "A" is consumed (used) by component "B".

Now the test of component "B" is failing because the alias seems not to be picked up. Notice, that I did not change a single line in component "B" at all (neither template, nor test etc.):

image

satanTime:
Why:

- #9684 reports that `MockComponents` loses the public alias of an input signal, causing Angular to reject the unchanged template binding.
- The behavior is fixed by #14490, but the existing coverage does not reproduce the `TestBed` declarations and `MockComponents` path from the report.

What:

- Add a focused regression that mocks the child through `MockComponents` and binds its signal input through the public alias.
- Assert that the mocked field remains a callable signal with the bound value.

Where:

- Add `tests/issue-9684/test.spec.ts`.
- Gate the spec to Angular targets with signal-input support in `test-spread.conf`.

Supersedes #14498.

Closes #9684

### DIFF
--- a/test-spread.conf
+++ b/test-spread.conf
@@ -63,6 +63,7 @@ file|tests/issue-12725/test.spec.ts|versions=21+
 file|tests/issue-13671/test.spec.ts|features=signalInputs
 file|tests/issue-14003/test.spec.ts|features=signalInputs
 file|tests/issue-8887/test.spec.ts|features=signalInputs
+file|tests/issue-9684/test.spec.ts|features=signalInputs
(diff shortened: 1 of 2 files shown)


satanTime:

PR #14501, opened 2026-08-02
fix(TestBed): preserve unmocked standalone imports #8649

### LINKED ISSUE #8649: Bug: MockComponent  (Standalone) mocks all components not just the ones specified
### Description of the bug

Using MockComponent to mock one dependent child Component all other child Components are mocked as well.

### An example of the bug

If you have total 3 components (All Components are **Standalone**):
MyComp  
MyComp1  
MyComp2   
And you want to Test MyComp but mock MyComp1 and keep the real MyComp2. 
The following example will result in all Components (MyComp1 & MyComp2) to get Mocked for some reason.

````html
<app-my-comp1></app-my-comp1>
<app-my-comp2></app-my-comp2>
````

````ts
describe('MyComp', () => {
 beforeEach(async () => {
    await TestBed.configureTestingModule({
      declarations: [
         MockComponent(MyComp1)
      ],
      imports: [
          MyComp
      ]
    }).compileComponents();
   ....
});
````

After a lot of try and error I figured out how to solve this.  
Workaround: 
````ts
await TestBed.configureTestingModule({
  declarations: [
    MockComponent(MyComp1),
    MyComp2
  ],
  imports: [
      MyComp
  ]
}).compileComponents();
````
By adding the MyComp2 to the declarations everything works as expected.   
I also tried to move declarations into the imports section like it's suggested for Standalone components, but this didn't make any difference.

Angular: 17.1.2 (Standalone Components)
NgMocks: 14.12.1

I would expect that only the specified Components that I put into ``MockComponent`` are replaced with Mock variants not everything else?

satanTime:
## Why

When a real standalone component and an explicitly mocked standalone child share an import, MockComponent can resolve a mock for that shared dependency before the mixed TestBed bridge rebuilds the module. The cached mock then replaces the dependency in the real target component even though the caller did not request it.

## What

- Preserve the recursive dependency graph of real TestBed declarations and imports while rebuilding mixed real and mock configurations.
- Keep explicit TestBed mocks and global ng-mocks resolutions higher priority than inferred real dependencies.
- Add cross-version regression coverage for declarations and imports usage, explicit and global mocking, and MockBuilder shallow rendering.

## Impact

Mixed TestBed configurations now mock only the standalone components requested by the caller. Shared imports of real roots remain real by default, while explicit mocks, global resolutions, and MockBuilder shallow behavior remain unchanged.

Fixes #8649

### DIFF
--- a/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts
+++ b/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts
@@ -17,6 +17,7 @@ import coreReflectMeta from './core.reflect.meta';
 import coreReflectProvidedIn from './core.reflect.provided-in';
 import { NG_MOCKS, NG_MOCKS_ROOT_PROVIDERS, NG_MOCKS_TOUCHES } from './core.tokens';
 import { AnyType, dependencyKeys } from './core.types';
+import { funcExtractDeps } from './func.extract-deps';
 import { getSourceOfMock } from './func.get-source-of-mock';
 import funcGetType from './func.get-type';
 import { isMockNgDef } from './func.is-mock-ng-def';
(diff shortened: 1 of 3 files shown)


satanTime:

PR #14504, opened 2026-08-02
fix(MockBuilder): support kept functional resolvers #7495

### LINKED ISSUE #7495: Bug: investigate the issue
https://stackoverflow.com/questions/77329862/ngmocks-error-ng0204-cant-resolve-all-parameters-for-pinresolver

satanTime:
## Why

When a functional resolver was passed to `MockBuilder.keep(...)` alongside routes supplied through `provideRouter`, `MockBuilder` also registered the resolver function as a root class provider. Angular then tried to construct the callback through dependency injection, treated its `ActivatedRouteSnapshot` and `RouterStateSnapshot` arguments as constructor dependencies, and failed with NG0204 before navigation.

## What

- Keep plain callback definitions without registering them as root class providers.
- Preserve provider registration for constructable classes and Angular factory providers.
- Add regression coverage for standalone and NgModule routing, including executable resolver arguments.
- Document class resolvers for Angular 5-22 and functional and standalone resolver setups for Angular 14-22.

## Impact

Kept functional resolvers can now remain in route definitions without Angular attempting to instantiate them as classes. Class-based resolvers and Angular factory providers retain their existing dependency-injection behavior across the supported Angular 5-22 range.

Fixes #7495


satanTime:

PR #14505, opened 2026-08-02
test(MockBuilder): cover nested module replacement #4486

### LINKED ISSUE #4486: Bug: "Nested" module replacement not working as expected
### Description of the bug

When using `MockBuilder` to replace a module import with a "testing module", that in turn imports some other "testing module" is not working as expected. The declarations of the sub-testing-module is not made available in my tests

### An example of the bug

Link: ﻿https://stackblitz.com/edit/github-fvnrfv?file=src/test.spec.ts

Expected behaviour: The component in the test renders as `name: sandboxt one testing two`

Actual behaviour: The component in the test renders as `name: sandboxt one two`, indicating the nested module declarations are not used

Brief of the code structure:
Component imports `Sub1Module` which declares `Component1` and imports `Sub2Module`. `Sub2Module` declares `Component2`.

In test, `Sub1Module` is replaced with `Sub1TestingModule`, which declares `Component1` but imports `Sub2TestingModule`. `Sub2TestingModule` declares `ComponentTesting2`.

The selectors of `Component2` and `ComponentTesting2` are identical.

We've been using this pattern for regular non-standalone component testing, where we usually provide all module imports manually to the test. We've started using MockBuilder for standalone component testing, since it's very convenient - but this one issue is a pretty big deal to us :(

satanTime:
## Why

- Replacing a module imported by a standalone target can change the transitive scope of a reused nested component.
- The existing parent-module ordering handles this scenario, but issue #4486 has no regression coverage to prevent it from returning.

## What

- Add a `MockBuilder.replace` regression that requires the nested component to render the replacement module's dependency.
- Spread the regression only to standalone-compatible Angular targets.

## Where

- `tests/issue-4486/test.spec.ts`
- `test-spread.conf`

Closes #4486

### DIFF
--- a/test-spread.conf
+++ b/test-spread.conf
@@ -69,6 +69,7 @@ file|tests/issue-7976/test.spec.ts|features=signalInputs
 file|tests/issue-8887/test.spec.ts|features=signalInputs
 file|tests/issue-9684/test.spec.ts|features=signalInputs
 file|tests/issue-8649/*.ts|features=standalone
+file|tests/issue-4486/*.ts|features=standalone
 file|tests/issue-6143/*.ts|versions=15+
 file|tests/issue-7938/*.ts|versions=15+
 file|tests/issue-11324/test.spec.ts|versions=15+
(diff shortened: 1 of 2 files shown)


satanTime:

PR #14506, opened 2026-08-02
test(TestBed): cover faster provider teardown #5437

### LINKED ISSUE #5437: Potential memory leak?
### Discussed in https://github.com/help-me-mom/ng-mocks/discussions/5435

<div type='discussions-op-text'>

<sup>Originally posted by **maxs-rose** April 12, 2023</sup>
After running ~2K tests the JS heap reaches 1GB of memory usage and no longer goes down.

image

There appears to be a lot of detached dom nodes handing around that seem to be related to `ngMocksUniverse`
image

Most of the test are setup similarly to the following:
```ts
describe("some component", () => {
     ngMocks.faster();

     const someServiceMock = MockService(SomeService);

     beforeAll(() => {
         return MockBuilder(SomeComponent).mock(SomeService, someServiceMock);
     );
     
    it("should do something", () => {
         const component = MockRender(SomeComponent).point.componentInstance;
         
         component.foo();

         expect(someServiceMock.serviceMethod).toHaveBeenCalled();
    });

    /* some more it blocks */
);
```

We have tried removing `ngMocks.faster` but that doesn't seem to have any affect.

Is this behavior due to how we are using NgMocks?</div>

satanTime:
Why:

- `ngMocks.faster()` previously discarded the TestBed module reference without destroying its injector, so kept providers did not receive `ngOnDestroy` between tests.
- The lifecycle gap reported in #5437 was fixed by #10825, but the existing unit coverage only verifies that `tearDownTestingModule()` is invoked.

What:

- Add an independent issue regression that renders a kept root provider and verifies its real `ngOnDestroy` lifecycle between tests.
- Run the regression on Angular 13+, where TestBed teardown is enabled by default.

Where:

- `tests/issue-5437/test.spec.ts`
- `test-spread.conf`

Closes #5437

### DIFF
--- a/test-spread.conf
+++ b/test-spread.conf
@@ -59,6 +59,7 @@ file|tests/issue-1596/test.spec.ts|versions=5-21
 file|tests/issue-296/test.spec.ts|versions=5-21
 file|tests/issue-736/test.spec.ts|versions=5-21
 file|tests/issue-7490/*.ts|versions=17+
+file|tests/issue-5437/*.ts|versions=13+
 file|tests/issue-10942/test.spec.ts|features=signalInputs
 file|tests/issue-11001/test.spec.ts|features=signalInputs
 file|tests/issue-13089/test.spec.ts|features=signalForms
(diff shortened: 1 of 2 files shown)


satanTime:

PR #14507, opened 2026-08-02
fix(MockComponent): handle circular standalone imports #8627

### LINKED ISSUE #8627: Bug: try to avoid recursive parsing when a standalone component imports itself
https://github.com/zyfra/Prizm/issues/1535

satanTime:
## Why

When `MockComponent` decorates a standalone declaration, it resolves the declaration's imports before publishing the completed mock. A direct self-import or a longer import cycle can therefore re-enter mock construction for a declaration already being decorated and recurse until the stack overflows.

## What

- Track declarations currently being constructed separately from completed mock and provider resolutions.
- Treat imports that point back into the active construction path as back-edges and omit them from generated mock metadata without caching the omission.
- Clean up construction state on both successful and exceptional paths.
- Add standalone regression coverage for a direct self-import and a two-component cycle.

## Impact

Standalone components with direct or transitive circular imports can now be mocked without recursive parsing. Completed mock caching, provider resolution, and subsequent independent mock construction retain their existing behavior.

Fixes #8627

### DIFF
--- a/libs/ng-mocks/src/lib/mock-module/create-resolvers.ts
+++ b/libs/ng-mocks/src/lib/mock-module/create-resolvers.ts
@@ -73,6 +73,14 @@ const createResolve =
     if (resolutions.has(def)) {
       return createResolveExisting(def, resolutions, change);
     }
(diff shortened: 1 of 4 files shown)


satanTime:

PR #14514, opened 2026-08-03
fix(MockService): harden function detection

satanTime:
## Why

`MockService` treated every function whose name started with `class` as a downleveled anonymous class, so ordinary callbacks such as `classify` were mocked as objects. Its webpack ES5 heuristic also interpolated function names directly into a regular expression, so valid names containing `$` were interpreted as regexp syntax and mocked as functions instead of classes.

## What

- Restrict anonymous downlevel class detection to TypeScript's `class_<number>` naming convention.
- Escape dynamic function names before matching webpack-style ES5 class calls and remove unnecessary regexp flags.
- Add internal and public `MockService` regressions for class-prefixed functions and `$` identifiers across the Angular spread.
- Correct the crawl test fixture's whitespace regular expression.

## Impact

`MockService` now distinguishes valid user functions from downleveled classes without changing normal class, function, Angular-decorated, or compiler-generated anonymous-class behavior.

### DIFF
--- a/libs/ng-mocks/src/lib/mock-service/check.is-func.spec.ts
+++ b/libs/ng-mocks/src/lib/mock-service/check.is-func.spec.ts
@@ -27,10 +27,18 @@ describe('check.is-func', () => {
 
   it('detects downleveled unnamed classes', () => {
     expect(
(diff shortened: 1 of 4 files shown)


satanTime:

PR #14515, opened 2026-08-03
fix(core): preserve defer blocks with mocks #7742

### LINKED ISSUE #7742: Bug: fixture.getDeferBlocks() returns no defer blocks if we want to use any ng-mocks mock inside the test
### Description of the bug

There a problem when trying to use ng-mocks in test which would like to use tools for new @defer block testing,
the problem is:
`fixture.getDeferBlocks();`
will return no defer blocks if we have any ng-mocks mock inside the test.

### An example of the bug

Here is minimalistic example to reproduce 

[angular-jest-defer-ng-mock.zip](https://github.com/help-me-mom/ng-mocks/files/13662909/angular-jest-defer-ng-mock.zip)

check `test-defer.component.spec.ts` in the example project

if we comment out `.overrideComponent(...)` from component ts file and `<test-child-defer />` line from template, the test will pass.

Mocking components with ng-mock has no effect on fixture.getDeferBlocks() functioning, it will return defer block correctly.

satanTime:
## Why

ng-mocks patches injected `ViewContainerRef` instances, but modern Angular already exposes the `__NG_ELEMENT_ID__` hook needed to intercept those injections. Wrapping `TestBed.createComponent` as well eagerly injected the root `ViewContainerRef`, which made `ComponentFixture.getDeferBlocks()` return an empty list on Angular 17 through 19.

The generated ESM bundle also declared Webpack runtime exports with top-level lexical bindings. Angular 14 can concatenate the package source more than once into one scope, causing `Identifier '__webpack_exports__' has already been declared` in its ES2015 build.

## What

- Use the `ViewContainerRef.__NG_ELEMENT_ID__` interception path on modern Angular and reserve the `TestBed.createComponent` wrapper for legacy Angular versions without that hook.
- Add an Angular 17+ regression that discovers and renders a defer block containing a mocked standalone dependency.
- Emit Webpack-generated ESM runtime bindings as `var` so Angular 14 can safely concatenate repeated package sources without changing the module format.

## Impact

`ComponentFixture.getDeferBlocks()` keeps returning the expected defer blocks when ng-mocks mocks are present, while dynamic `ViewContainerRef` operations remain patched. The ESM package also loads without duplicate `__webpack_exports__` declarations in Angular 14 ES2015 builds.

## Validation

- Root coverage: 1,737 passed, 2 skipped; 100% statements, branches, functions, and lines
- Angular 5 ES5: 1,228 browser tests; 393 Jest suites
- Angular 14 ES5 and ES2015: 1,339 browser tests and 411 Jest suites in each mode
- Angular 17: 1,365 browser tests; 423 Jest suites
- Angular 18: 1,365 browser tests; 423 Jest suites
- Angular 19 after rebase: 1,343 browser tests; 425 Jest suites
- Angular 20 after rebase: 1,345 browser tests; 426 Jest suites
- Lint and TypeScript checks

Closes #7742

### DIFF
--- a/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts
+++ b/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts
@@ -376,8 +376,10 @@ const patchVcrInstance = (vcrInstance: ViewContainerRef) => {
   }
 };
 
+// istanbul ignore next: legacy Angular lacks __NG_ELEMENT_ID__; covered by the compatibility matrix.
 const createComponent =
   (original: TestBedStatic['createComponent'], instance: TestBedStatic): TestBedStatic['createComponent'] =>
+  // istanbul ignore next: legacy Angular lacks __NG_ELEMENT_ID__; covered by the compatibility matrix.
   (...args) => {
     const fixture = original.call(instance, ...args);
     try {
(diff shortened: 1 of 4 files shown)


satanTime:

Issue #10919, opened 2025-01-29
Bug: AutoSpy not working as expected


EinfachHans:
### Description of the bug

I have an Angular service (generated by an Open Api Swagger file), which looks like this:

```ts
@Injectable({ providedIn: 'root' })
export class MyService {
  info = {
    request: () => {
      return ...
    }
  }
}
```

The requests are mapped into objects if they are defined in the swagger file like that.

The `autoSpy` of this (awesome btw) library seems to always mocks objects to `undefined`? 

What i would expect would be the object to exist with a function `request` which should be the mocked function.

With the current implementation it is not possible for me to use it correctly, because i receive the error:

`TypeError: Cannot read properties of undefined (reading 'request'`) when calling `this.service.info.request()`

satanTime:
[v14.16.0](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.16.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14554, opened 2026-08-05
fix(MockService): suppress class constructors #14528

### LINKED ISSUE #14528: Bug: `MockProvider` no longer suppresses the real constructor
### Description of the bug

`MockProvider(Service, partialOverrides)` no longer suppresses the real constructor of `Service`. It appears to have regressed in `14.16.0`.

The docs for `MockService`/`MockProvider` say the result "does nothing, it's dummy" (held through in `14.15.3`). In `14.16.0`, if the real class has a constructor with a side effect, that side effect now runs.

### An example of the bug

```ts
import { MockProvider } from 'ng-mocks';

const constructorSpy = jasmine.createSpy('realConstructor');

class SideEffectService {
    constructor() {
        constructorSpy(); // stands in for e.g. subscribing + firing an HTTP request
    }

    doThing(): string {
        return 'real';
    }
}

it('must not run the real constructor of a service mocked via MockProvider', () => {
        providers: [MockProvider(SideEffectService, { doThing: () => 'mocked' })],
    });

    const instance = TestBed.inject(SideEffectService);

    expect(instance.doThing()).toEqual('mocked');
    expect(constructorSpy).not.toHaveBeenCalled(); // passes on 14.15.3, fails on 14.16.0
});
```

`MockProvider`/`MockService` should continue to never execute the mocked class's real constructor, even when only some members are overridden and even when the class is `providedIn: 'root'`.

satanTime:
## Why

`MockService` instantiated classes whose reported constructor arity was zero to discover own object properties. Because `MockProvider` delegates to `MockService`, real constructor side effects ran even when provider behavior was overridden. `Function.length` also treats default-parameter constructors as zero-argument.

Instance fields do not exist until construction. Automatically discovering them would require either executing application code or parsing compiled source, neither of which is safe across supported compilation targets.

## What

- Restore prototype-only class mocking so user constructors are never invoked.
- Require own-property shapes explicitly through the existing `MockService` and `MockProvider` overrides, retaining nested auto-spies by mocking the supplied plain shape.
- Add regression coverage for side-effectful no-argument and default-parameter constructors, including the reported `MockProvider` flow.
- Update the #10919 coverage and API docs for the explicit-shape contract.

## Impact

`MockProvider` and `MockService` no longer execute application constructors and behave as they did before the own-property discovery change. Prototype methods remain automatic mocks; callers that need constructor-created fields provide their shape explicitly.

Fixes #14528


satanTime:

PR #14557, opened 2026-08-05
fix(MockBuilder): mock inject dependencies in kept services #14544

### LINKED ISSUE #14544: Bug: MockBuilder does not mock inject()-ed dependencies when the kept declaration is a service (follow-up to #9397)
### Description of the bug

`#9397` ("MockBuilder: standalone inject services are properly mocked", released in `v14.16.0`) fixed dependency discovery for `inject()` calls in **standalone declarations**. The same discovery still does not happen when the class passed to `MockBuilder` as the `keep` target is a plain `@Injectable` (or Angular 22's `@Service()`) class.

For a service, `MockBuilder` appears to resolve dependencies from constructor parameter metadata only, so a dependency assigned via a field initialiser (`private readonly api = inject(ApiService)`) is never registered as a mock provider and the real instance is constructed.

In a trivial repro this shows up as the dependency simply not being mocked. In a real codebase the usual symptom is a hard failure, because the real dependency pulls in its own unresolvable providers:

```
NG0201: No provider found for `InjectionToken httpConfig`.
Source: DynamicTestModule. Path: BasisInfoService -> BasisinfoApiService -> CustomHttpClient -> InjectionToken httpConfig
```

This makes the standard `@Injectable` + constructor → `@Service()` + `inject()` migration a breaking change for every service spec that relies on `MockBuilder` auto-mocking.

Versions: `ng-mocks@14.16.0`, `@angular/core@22.1.0`, `jest@30.4.2`, `jest-preset-angular@17.0.0` (zoneless test env).

### An example of the bug

Link:

Paste into `src/test.spec.ts` of the sandbox template. `A` and `C` pass, `B` fails:

```ts
import { Component, inject, Injectable } from '@angular/core';
import { MockBuilder, MockRender } from 'ng-mocks';

class DepService {
  public ping(): string {
    return 'real';
  }
}

// A — standalone component using inject(): fixed by #9397
@Component({ selector: 'app-probe', standalone: true, template: '' })
class ProbeComponent {
  public readonly dep = inject(DepService);
}

// B — service using inject(): still not mocked
class ProbeService {
  public readonly dep = inject(DepService);
}

// C — service using constructor injection: works
class ProbeCtorService {
  public constructor(public readonly dep: DepService) {}
}

describe('MockBuilder + inject() in a service', () => {
  describe('A: standalone component + inject()', () => {
    beforeEach(() => MockBuilder(ProbeComponent));

    it('mocks the dependency', () => {
      const dep = MockRender(ProbeComponent).point.componentInstance.dep;
      expect(dep.ping()).toBeUndefined(); // passes
    });
  });

  describe('B: service + inject()', () => {
    beforeEach(() => MockBuilder(ProbeService));

    it('mocks the dependency', () => {
      const dep = MockRender(ProbeService).point.componentInstance.dep;
      expect(dep.ping()).toBeUndefined(); // FAILS — returns 'real'
    });
  });

  describe('C: service + constructor injection', () => {
    beforeEach(() => MockBuilder(ProbeCtorService));

    it('mocks the dependency', () => {
      const dep = MockRender(ProbeCtorService).point.componentInstance.dep;
      expect(dep.ping()).toBeUndefined(); // passes
    });
  });
});
```

**Expected:** `MockBuilder(ProbeService)` discovers and mocks `DepService`, exactly as it does for a standalone component (`A`) and for constructor injection (`C`). `dep.ping()` returns `undefined` and `jest.isMockFunction(dep.ping)` is `true`.

**Actual:** `DepService` is instantiated for real. `dep.ping()` returns `'real'` and `jest.isMockFunction(dep.ping)` is `false`. The only workaround is to name every `inject()`-ed dependency explicitly: `MockBuilder(ProbeService).mock(DepService)`.

satanTime:
## Why

`MockBuilder(Service)` mocked constructor-injected root dependencies, but a migration to field `inject()` resolved the real dependency instead. The runtime injection support added for standalone declarations only wrapped declaration factories, so kept service factories still ran outside that scoped mocking window.

An earlier environment initializer could also instantiate the kept service before the ng-mocks initializer installed the runtime hook.

The repository guidance did not distinguish detailed review context in pull request descriptions from commitlint's 100-character limit for commit-message lines.

## What

- Wrap the exact provider created by `MockBuilder` for a directly kept service in the runtime injection scope while preserving its factory and dependencies.
- Install the injector hook lazily from the service factory, keep the environment initializer only for standalone declarations, and use one injector registry for idempotent setup and cleanup.
- Add regressions for constructor parity, root and local `Injectable` services, eager environment initialization, Angular 22 `Service`, and kept factory-provider dependencies.
- Clarify that pull request descriptions should retain complete `Why`, `What`, and `Impact` or `Where` context, while every commit-message line stays within 100 characters.

## Impact

`MockBuilder(Service)` continues to mock service dependencies after constructor injection is migrated to `inject()`. Real dependency constructors no longer run in that path, while explicit keeps, custom provider factories, dependency ordering, and the standalone declaration behavior remain intact.

Future changes can keep their pull request explanation complete without violating commitlint constraints.

Fixes #14544


satanTime:

PR #14558, opened 2026-08-05
fix(build): restore CommonJS Angular initialization order

### LINKED ISSUE #14545: Bug: 14.16.0: reordered CJS externals break partially-compiled @angular/common ("_PlatformLocation needs to be compiled using the JIT compiler")
### Description of the bug

`v14.16.0` changed the order of the external `require()` calls in the CJS/UMD entry (`index.js`). `@angular/common` moved from fourth position to first:

```js
// 14.15.3
require("@angular/core"), require("@angular/core/testing"), require("@angular/forms"), require("@angular/common"), require("@angular/platform-browser")

// 14.16.0
require("@angular/common"), require("@angular/core"), require("@angular/core/testing"), require("@angular/forms"), require("@angular/platform-browser")
```

`@angular/common` is partially compiled and eagerly evaluates `_PlatformLocation` at module scope (`@angular/common/fesm2022/_platform_location-chunk.mjs`, the only `ɵɵngDeclareInjectable` in the package), which requires the JIT compiler facade to be registered.

Previously `@angular/core/testing` was required first, and it does `import { ResourceLoader } from '@angular/compiler'` (`@angular/core/fesm2022/testing.mjs:11`), which registers the facade as a side effect. With the new order, `@angular/common` is evaluated before anything has loaded `@angular/compiler`, and merely importing `ng-mocks` throws:

```
The injectable '_PlatformLocation' needs to be compiled using the JIT compiler,
but '@angular/compiler' is not available.

  at Object.ɵɵngDeclareFactory (@angular/core/fesm2022/core.mjs:3087:20)
  at Object.&lt;anonymous&gt; (@angular/common/fesm2022/_platform_location-chunk.mjs:23:20)
  at Object.&lt;anonymous&gt; (@angular/common/fesm2022/common.mjs:7:66)
  at node_modules/ng-mocks/index.js:1:86
  at Object.&lt;anonymous&gt; (test-setup.ts:23:1)   // import { ngMocks } from 'ng-mocks'
```

This surfaces as `Test suite failed to run` for every spec.

It only bites setups that do not otherwise load `@angular/compiler`. Notably `jest-preset-angular`'s zoneless environment (`jest-preset-angular/setup-env/zoneless`) bootstraps via `platformBrowserTesting` from `@angular/platform-browser/testing` rather than `@angular/platform-browser-dynamic`, so it never pulls the compiler in. Zone-based setups are unaffected because they load it through the dynamic platform.

Arguably ng-mocks was only ever working here by accident, and the real fix belongs in the consuming setup. But since the ordering change alone flips a whole class of projects from green to "suite failed to run" on a minor bump, it seems worth either restoring the previous order or documenting the requirement.

Versions: `ng-mocks@14.16.0` (works on `14.15.3`), `@angular/core@22.1.0`, `@angular/common@22.1.0`, `jest@30.4.2`, `jest-preset-angular@17.0.0` with `createCjsPreset()` + `setupZonelessTestEnv()`.

### An example of the bug

Link:

Not reproducible in the sandbox templates, as those use the zone-based setup which loads `@angular/compiler` via `@angular/platform-browser-dynamic`. To reproduce:

1. Any Angular 22 + Jest project using `jest-preset-angular@17`'s `createCjsPreset()`.
2. Bootstrap the test env with `setupZonelessTestEnv()` from `jest-preset-angular/setup-env/zoneless`.
3. Ensure nothing imports `@angular/compiler` or `@angular/core/testing` before `ng-mocks`.
4. `import { ngMocks } from 'ng-mocks';` in a setup file → throws on `14.16.0`, passes on `14.15.3`.

**Expected:** Importing `ng-mocks` does not depend on the host project having already registered the JIT compiler facade — a patch/minor bump should not change which Angular entry points get evaluated first.

**Actual:** Importing `ng-mocks` evaluates `@angular/common` before `@angular/core/testing`, so partially-compiled injectables in `@angular/common` fail to link and every suite fails to run. Workaround is to add `import '@angular/compiler';` as the first import of the Jest setup file.

satanTime:
Closes #14545

Why:

- Webpack 5.109 changed the generated UMD external order, allowing `@angular/common` to evaluate before `@angular/core/testing` registers Angular's JIT facade.
- Zoneless Jest consumers that do not otherwise load `@angular/compiler` then fail while importing `ng-mocks`.

What:

- Preload `@angular/core/testing` only on the CommonJS path before the UMD wrapper resolves Angular externals.
- Fail the webpack build if the final CommonJS bundle does not load `@angular/core/testing` before `@angular/common`.

Where:

- `webpack.config.js`

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,5 +1,29 @@
 const path = require('node:path');
 
+const { BannerPlugin, Compilation } = require('webpack');
+
+class AngularCommonJsOrderPlugin {
+  apply(compiler) {
+    compiler.hooks.thisCompilation.tap(AngularCommonJsOrderPlugin.name, compilation => {
+      compilation.hooks.processAssets.tap(
+        {
+          name: AngularCommonJsOrderPlugin.name,
+          stage: Compilation.PROCESS_ASSETS_STAGE_REPORT,
+        },
+        () => {
+          const source = compilation.getAsset('index.js')?.source.source().toString() ?? '';
+          const coreTestingIndex = source.indexOf('@angular/core/testing');
+          const commonIndex = source.indexOf('@angular/common');
+
+          if (coreTestingIndex === -1 || commonIndex === -1 || coreTestingIndex > commonIndex) {
+            throw new Error('index.js must load @angular/core/testing before @angular/common');
+          }
+        },
+      );
+    });
+  }
+}
+
 const performance = {
   hints: 'error',
   maxAssetSize: 800 * 1024,
@@ -22,6 +46,15 @@ module.exports = [
     },
     externals: /^@angular\//,
     performance,
+    plugins: [
+      // The UMD external order is not stable across webpack versions. In CommonJS,
+      // core/testing must register Angular's JIT facade before common is evaluated.
+      new BannerPlugin({
+        banner: "if (typeof exports === 'object' && typeof module === 'object') require('@angular/core/testing');",
+        raw: true,
+      }),
+      new AngularCommonJsOrderPlugin(),
+    ],
     module: {
       rules: [
         {


satanTime:

PR #14559, opened 2026-08-06
feat(MockRender): support signal input initial values #11101

### LINKED ISSUE #11101: Feature Request:  Support for InitialValues of Signal Inputs

It seems like the ng-mocks library does not automatically apply default values to signal inputs when rendering components for testing. This means that even if a signal input has a default value set in the component (e.g., input<T>(defaultValue)), it appears to be null in the test environment unless manually set. This behavior might not accurately reflect how the component would work in a real application, especially if the default value is removed from the component.

### Proposed solution
If this behavior is indeed due to how ng-mocks handles signal inputs, it would be beneficial to implement a feature that automatically applies default values to signal inputs when rendering components for testing:

- Tests would automatically fail if default values are removed from the component, ensuring that changes to default values are caught and it mimicks the component's behavior better.

satanTime:
## Why

`MockRender(Component)` without params intentionally binds inputs to `null`. Applying that behavior to optional signal inputs discards `input(initialValue)`, so tests cannot observe the component default and must pass empty params as a workaround.

## What

- Preserve initialized signal input values in the generated `MockRender` wrapper.
- Keep aliases, required signals, input transforms, and later `setInput` updates consistent with Angular.
- Add focused source coverage and an Angular 17+ compatibility regression.
- Clarify the documented no-params behavior.
- Record clean Angular CLI cache and Docker recovery guardrails in the AI issue-triage guidance.

## Impact

`MockRender(Component)` without params now mirrors Angular component construction for optional signal input defaults. Ordinary inputs and required signal inputs retain their existing null-binding behavior. No public API is added.

Closes #11101

### DIFF
--- a/.agents/skills/triage-issue/SKILL.md
+++ b/.agents/skills/triage-issue/SKILL.md
@@ -16,7 +16,7 @@ Create and maintain a plain Markdown checklist:
 - [ ] Create a dedicated issue worktree from `upstream/main`
 - [ ] Reproduce the bug with a local `issue-*` regression test
 - [ ] Fix the implementation without changing the reproducer test
-- [ ] Run coverage and affected e2e validation
+- [ ] Clear affected Angular CLI caches and run coverage and e2e validation
 - [ ] Prepare comments, commit message, and PR against `upstream/main`
 ```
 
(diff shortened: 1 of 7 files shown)


satanTime:

Issue #11099, opened 2025-02-27
Bug: Zoneless test


handja:
I've activated Zoneless on my project, but now I encounter an issue on unit test with ng-mocks (issue not encountered with TestBed)

Issue : **Error: NG0908: In this configuration Angular requires Zone.js**

TS FILE
```
import { Component, input } from '@angular/core';

@Component({
  selector: 'esac-test',
  template: '',
})
export class TestComponent {
  readonly count = input(0);
}
```

SPEC FILE
```
import { provideExperimentalZonelessChangeDetection } from '@angular/core';
import { TestComponent } from './control.component';
import { MockBuilder, MockedComponentFixture, MockRender } from 'ng-mocks';

fdescribe('TestComponent', () => {
  let component: TestComponent;
  let fixture: MockedComponentFixture<TestComponent>;

  beforeEach(async () => {
    await MockBuilder(TestComponent).provide(provideExperimentalZonelessChangeDetection());
    fixture = MockRender(TestComponent);
    component = fixture.point.componentInstance;
  });

  it('should create', () => {
    fixture.componentRef.setInput('count', 1);
    expect(component).toBeTruthy();
  });
});

```

GipHub123:
@handja I haven't had issues with zoneless test cases.  At least this work

test.ts
```
@NgModule({
  providers: [
    provideZonelessChangeDetection()
  ]
})
export class TestingZonelessModule {
}

getTestBed().initTestEnvironment(
  [BrowserTestingModule, TestingZonelessModule],
  platformBrowserTesting(),
  {
    teardown: { destroyAfterEach: false }
  }
);
```
angular.json
```
{
  "test": {
    "builder": "@angular-devkit/build-angular:karma",
    "options": {
      "main": "src/test.ts"
    }
  }
}
```

GipHub123:
@handja Is this still an issue? If not, could you please close it?

handja:
Sorry, I didn't use it anymore, I can close the issue if it works on your case !

GipHub123:
@handja Please close it 👍 It helps keep the issue log cleaner and makes it easier to focus on the relevant issues.

Issue #7802, opened 2023-12-25
support vitest


satanTime:
https://github.com/Tallyb/ng-mocks-vitest @satanTime

_Originally posted by @Tallyb in https://github.com/help-me-mom/ng-mocks/discussions/5064#discussioncomment-7876580_

PR #14563, opened 2026-08-07
fix(MockRender): schedule zoneless input updates #12061

### LINKED ISSUE #12061: Bug: Can't change inputs of components with zoneless change detection
### Description of the bug

I'm using Angular 20.2 with vitest through analogjs plugin.
We recently made the switch to zoneless because its promoted stable.
When rendering Components through the `MockRenderFactory` and passing input values via a props object (like its mentioned in the docs), changing the props and run `fixture.detectChanges()` where sufficient to trigger changeDetection in zonefull.

With zoneless this doesn't work anymore. Tried with `fixture.whenStable()` and `fixture.whenRenderingDone()`, `ngMocks.input()` and `setInput` from the componentRef, but no luck. Also tried simple `MockRender`.

Only running `fixture.changeDetectorRef.markForCheck` will trigger change Detection.

When trying this with `TestBed.createComponent` it will work like expected when setting Input via `setInput`, so I guess this is related to how the MockRender works but idk. 

Maybe I'm doing something wrong

Thanks in advance

satanTime:
## Why

Issue #12061 shows that changing MockRender or MockRenderFactory params updates ng-mocks' source object without
notifying Angular's zoneless scheduler. OnPush targets can therefore remain stale until callers manually mark them
for check, even though the equivalent ComponentRef.setInput flow schedules correctly.

Issue #11099 previously documented the broader need for zoneless TestBed compatibility. Issue #7802 tracks Vitest
support separately; this change intentionally stays runner-neutral and adds no Vitest dependency or project.

## What

- Route writable MockRender input params through ComponentRef.setInput for zoneless fixtures while keeping the
  params object as the source of truth.
- Add an issue regression covering MockRender, MockRenderFactory, plain inputs, compiled signal inputs, and the
  direct TestBed baseline.
- Run the existing Angular 20, 21, and 22 version projects in zoned and zoneless Jasmine and Jest modes using the
  official Angular provider and jest-preset-angular setup entry point.
- Keep the Zone-dependent fakeAsync suites in their existing zoned coverage and exclude only those suites from the
  zoneless modes.

## Impact

Angular 20-22 tests can mutate MockRender params and let the zoneless fixture stabilize without a manual
markForCheck workaround. Zoned behavior is unchanged because the scheduling bridge is enabled only for zoneless
fixtures. Existing version projects and CI jobs are reused; no new dependencies, projects, or jobs are added.

Closes #12061.
Related to #11099.
Related to #7802.


satanTime:

Issue #9397, opened 2024-07-09
Bug: inject function can not be mocked


wellwind:
### Description of the bug

The auto spy feature can not mock instance injected by `inject` function.

### An example of the bug

service that want be mocked:

```typescript
import { Injectable } from '@angular/core';

@Injectable({ providedIn: 'root' })
export class TodoListService {
  someMethod() {}
}
```

the component:

```typescript
import { Component, inject } from '@angular/core';
import { TodoListService } from './todo-list.service';

@Component({
  selector: 'app-root',
  standalone: true,
  template: ``,
})
export class AppComponent {
  // can not be auto spied
  public todoListService = inject(TodoListService);

  // can be auto spied
  // constructor(private todoListService: TodoListService) {}

  someMethod() {
    this.todoListService.someMethod();
  }
}
```

test code:

```typescript
import { TestBed } from '@angular/core/testing';
import { MockBuilder, MockRender, ngMocks } from 'ng-mocks';
import { AppComponent } from './app.component';
import { TodoListService } from './todo-list.service';

ngMocks.autoSpy('jasmine');

describe('AppComponent', () => {
  let component: AppComponent;

  beforeEach(() => MockBuilder(AppComponent));

  beforeEach(() => {
    component = MockRender(AppComponent).componentInstance;
  });

  it('should call call todoListService.someMethod', () => {
    const todoListService = TestBed.inject(TodoListService);

    component.someMethod();
    expect(todoListService.someMethod).toHaveBeenCalled();
  });
});
```

Link:

https://github.com/wellwind/ng-mocks-inject-demo

### Expected vs actual behavior

In component, when I use ctor to inject service, it works fine.

Bue when I use `inject` function, I got errors:

>        Error: <toHaveBeenCalled> : Expected a spy, but got Function.
>        Usage: expect(<spyObj>).toHaveBeenCalled()

However, it should also passed the test.

ste2425:
If it helps, manually providing with `.provide()` chained after the `MockBuilder` does work.

Well it works in that i can mock the service but cannot use `MockInstance` to change that mock.

Just FYI incase it helps investigate.

pcbowers:
[...]

PR #14564, opened 2026-08-07
fix(MockBuilder): mock inject dependencies in classic components #14560

### LINKED ISSUE #14560: Bug: MockBuilder does not auto-mock dependencies obtained via inject() in class field initializers / constructor body (only detects constructor parameters)
### Summary
`MockBuilder(Component)` analyzes the component's **constructor parameters** to determine which dependencies to auto-mock. Dependencies obtained through the functional `inject()` API (either as class field initializers or inside the constructor body) are **not** discovered and therefore not auto-mocked. This causes previously green tests to fail when a component is refactored from constructor-parameter DI to `inject()`, even though both patterns are semantically equivalent in Angular.

The Angular team has been actively pushing developers toward `inject()` (see the `@angular-eslint/prefer-inject` rule and the official `ng generate @angular/core:inject` migration schematic), so this gap causes real friction when modernizing codebases.

### Environment
- `ng-mocks`: 14.16.1
- `@angular/core`: 22.1.0
- `typescript`: 6.0.3
- TS target: `ES2022` (default `useDefineForClassFields: true`)

### Steps to reproduce

Given a service and a component:

```ts
export class LatinCharacterValidationService {
  containsOnlyLatin(value: string): boolean {
    return /^[\x00-\x7F]*$/.test(value);
  }
}
```

**Working version (constructor param DI):**
```ts
@Component({ selector: 'my-input', template: '' })
export class MyInputComponent {
  public constructor(private readonly service: LatinCharacterValidationService) {}

  public validate(value: string) {
    return this.service.containsOnlyLatin(value);
  }
}
```

**Refactored version (idiomatic Angular `inject()`):**
```ts
@Component({ selector: 'my-input', template: '' })
export class MyInputComponent {
  private readonly service = inject(LatinCharacterValidationService);

  public validate(value: string) {
    return this.service.containsOnlyLatin(value);
  }
}
```

Test:
```ts
describe('MyInputComponent', () => {
  beforeEach(() => MockBuilder(MyInputComponent));

  it('mocks the service', () => {
    const fixture = MockRender(MyInputComponent);
    const service = TestBed.inject(LatinCharacterValidationService) as Mocked<LatinCharacterValidationService>;

    service.containsOnlyLatin.mockReturnValue(true); // ← throws

    expect(fixture.point.componentInstance.validate('test')).toBe(true);
  });
});
```

### Expected behavior
Both versions of `MyInputComponent` should behave identically under `MockBuilder(MyInputComponent)` — `LatinCharacterValidationService` should be auto-mocked in both cases, since `inject()` and constructor-param DI are equivalent at runtime.

### Actual behavior
- **Constructor version:** `LatinCharacterValidationService` is auto-mocked; `service.containsOnlyLatin` is a jest mock, `mockReturnValue` works, test passes. ✅
- **`inject()` version:** `LatinCharacterValidationService` is resolved to the real implementation (since it's `providedIn: 'root'`). `service.containsOnlyLatin.mockReturnValue` throws:
  ```
  TypeError: service.containsOnlyLatin.mockReturnValue is not a function
  ```
  Test fails. ❌

### Workaround
Explicitly declare the dependency to be mocked:
```ts
beforeEach(() => MockBuilder(MyInputComponent).mock(LatinCharacterValidationService));
```

This works but requires tracking every field-`inject()` dependency by hand, negating one of the main ergonomic benefits of `MockBuilder`.

### Related
Looks like the same underlying cause as #14544.

### Impact
Any codebase adopting the Angular-recommended `inject()` pattern (or applying `ng g @angular/core:inject`) will silently see tests break — often with confusing errors like `mockReturnValue is not a function` — because a real service instance leaks into what the developer expects to be a mocked test environment. It would be great if `MockBuilder` could statically detect `inject()` calls (field initializers and constructor bodies) the same way it detects constructor parameters, or offer an opt-in mode that mocks all `providedIn: 'root'` services referenced by the component.

satanTime:
## Why

Runtime `inject()` support covered standalone declarations and was recently extended to kept services, but classic non-standalone components remained outside both factory hooks.

In one-argument `MockBuilder(Component)` auto-mock mode, moving a dependency from a constructor parameter to a field initializer or constructor-body `inject()` therefore resolved the real root service.

## What

- Enable the existing runtime injection hook for classic kept declarations in one-argument auto-mock mode.
- Preserve existing root-provider behavior for two-argument builders and explicitly kept modules.
- Add regressions for both field-initializer and constructor-body `inject()` calls.
- Spread the regression across Angular versions that expose the `inject()` API.

## Impact

Classic components retain MockBuilder auto-mocking after constructor injection is migrated to `inject()`. Standalone declarations, kept services, and kept-module root-provider semantics remain unchanged.

Related to #9397 and #14544.

Fixes #14560

### DIFF
--- a/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.ts
+++ b/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.ts
@@ -91,7 +91,12 @@ export class MockBuilderPromise implements IMockBuilder {
       handleEntryComponents(ngModule);
       applyPlatformModules();
 
-      const runtimeInjectProvider = createRuntimeInjectProvider(this.keepDef, this.configDef, ngModule.providers);
+      const runtimeInjectProvider = createRuntimeInjectProvider(
+        this.keepDef,
+        this.configDef,
+        ngModule.providers,
+        this.configDefault.dependency !== true,
+      );
       if (runtimeInjectProvider) {
         ngModule.providers.push(runtimeInjectProvider);
       }
--- a/libs/ng-mocks/src/lib/mock-builder/promise/create-runtime-inject-provider.spec.ts
+++ b/libs/ng-mocks/src/lib/mock-builder/promise/create-runtime-inject-provider.spec.ts
@@ -49,6 +49,7 @@ describe('create-runtime-inject-provider', () => {
         [TargetWithoutDependencies, { shallow: false }],
       ]),
       providers,
+      false,
     );
 
     const destroyCallbacks: Array<() => void> = [];
--- a/libs/ng-mocks/src/lib/mock-builder/promise/create-runtime-inject-provider.ts
+++ b/libs/ng-mocks/src/lib/mock-builder/promise/create-runtime-inject-provider.ts
@@ -7,16 +7,30 @@ import { isStandalone } from '../../common/func.is-standalone';
 import { installRuntimeInject, runRuntimeInject } from '../../common/ng-mocks-runtime-inject';
 import ngMocksUniverse from '../../common/ng-mocks-universe';
 
-export default (keepDef: Set<any>, configDef: Map<any, any>, providers: any[]): Provider | undefined => {
+export default (
+  keepDef: Set<any>,
+  configDef: Map<any, any>,
+  providers: any[],
+  autoMockRootProviders: boolean,
+): Provider | undefined => {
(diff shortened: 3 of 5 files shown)


satanTime:

Vinccool96:
It doesn't seem to have been published on NPM

Issue #1932, opened 2022-02-26
support for tree-shakable tokens


satanTime:
Now a14 is failing again.

https://github.com/angular/angular/releases/tag/14.0.0-next.4
https://github.com/angular/angular/pull/45102/files#diff-8f95458cfc36347826d78314a6d25460dbf4fd7e4f82f63ed5c95dc8d692c691R111

satanTime:
[v13.0.4](https://github.com/ike18t/ng-mocks/releases/tag/v13.0.4) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14614, opened 2026-08-11
fix(core): preserve Angular EffectScheduler #14613

### LINKED ISSUE #14613: Bug: Invalid EffectScheduler initially provided in Angular 22.
### Description of the bug

After upgrading from Angular 21 to Angular 22, tests using MockBuilder fail when a service/interceptor uses effect().

Error message:

`TypeError: node.scheduler.add is not a function`

The problem seems related to ng-mocks default root-provider mocking.

Minimal relevant setup:
Angular 22.1.1
ngMocks 14.15.1 or v14.17 (issue happens with both)

(the issue did not occur with Angular 21.2.18)

### An example of the bug

```ts
// setup-jest.ts
import '@testing-library/jest-dom';
import { ngMocks } from 'ng-mocks';
import { setupZonelessTestEnv } from 'jest-preset-angular/setup-env/zoneless';

ngMocks.autoSpy('jest');
setupZonelessTestEnv();
```

```ts
// my-interceptor.spec.ts
import { effect, Injectable } from '@angular/core';
import { MockBuilder, ngMocks } from 'ng-mocks';

class MyInterceptor {
  constructor() {
    effect(() => {
      console.log('hi');
    });
  }
}
describe('MyInterceptor', () => {
  beforeAll(() => MockBuilder(MyInterceptor));

  it('fails', () => {
    ngMocks.findInstance(MyInterceptor);
    expect(true).toBe(true);
  });

  it('pass', () => {
    ngMocks.findInstance(MyInterceptor);
    expect(true).toBe(true);
  });
});
```

Expected:
The instance is created without any errors.

Actual:
The error occurs for the first test, the second test passes (and any other tests added also pass).

If I only run the second test, it fails.

### Stuff I tried that fixes the issue

1. adding `it('noop', () => expect(true).toBe(true));` before other tests.
2. adding `.keep(NG_MOCKS_ROOT_PROVIDERS)` in the mockbuilder chain
3. adding `.keep(ɵEffectScheduler)` in the mockbuilder chain
4. adding `ngMocks.globalKeep(ɵEffectScheduler)` to setup-jest
5. adding `{ teardown: { destroyAfterEach: false } }` to the `setupZonelessTestEnv` function in setup-jest
6. adding `beforeEach(() => TestBed.configureTestingModule({ providers: [MyInterceptor] }));`

### Stuff I tried that does not fix the issue

1. Use beforeEach instead of beforeAll. When using beforeEach, all tests fail instead of just the first.
2. adding `ngMocks.globalKeep(NG_MOCKS_ROOT_PROVIDERS)` to setup-jest

### Other debugging I did

When I add `console.log(TestBed.inject(ɵEffectScheduler));` to the top of the tests, the output is
```
ZoneAwareEffectScheduler { dirtyEffectCount: 0, queues: Map(0) {} }
ZoneAwareEffectScheduler { dirtyEffectCount: 0, queues: Map(0) {} }
```

When I add `console.log('scheduler:', inject(ɵEffectScheduler));` to the constructor of the interceptor, the output is
```
scheduler: _EffectScheduler {}
scheduler: ZoneAwareEffectScheduler { dirtyEffectCount: 0, queues: Map(0) {} }
```
ie: the mockbuilder provides a different (invalid) effectscheduler, but only for the first test.

satanTime:
## Why

Angular 22 effects register themselves through the root `EffectScheduler`. `MockBuilder`
replaced that provider with an empty class mock, so the first effect creation failed when
Angular called the missing `add` method.

## What

- Preserve both Angular runtime names for `EffectScheduler` by default while keeping
  explicit `.mock(...)` overrides available.
- Add an Angular 22 regression test for a root service that creates an effect.
- Follow the framework root-provider preservation pattern introduced for #1932.

## Impact

Root injectables can create Angular 22 effects under `MockBuilder` in both zoned and
zoneless test profiles without requiring `NG_MOCKS_ROOT_PROVIDERS` workarounds.

Fixes #14613

### DIFF
--- a/libs/ng-mocks/src/lib/common/core.config.ts
+++ b/libs/ng-mocks/src/lib/common/core.config.ts
@@ -29,6 +29,9 @@ export default {
     'IterableDiffers',
     'KeyValueDiffers',
 
+    // Angular 22 effects require the concrete root scheduler implementation.
+    'EffectScheduler',
+
     // Angular 16 adds underscores
     '_DomRendererFactory2',
     '_EventManager',
@@ -41,6 +44,7 @@ export default {
     '_Compiler',
     '_IterableDiffers',
     '_KeyValueDiffers',
+    '_EffectScheduler',
   ],
   neverMockToken: [
     'InjectionToken Set Injector scope.', // INJECTOR_SCOPE // ivy only
--- a/test-spread.conf
+++ b/test-spread.conf
@@ -59,6 +59,7 @@ file|tests/issue-14528/*.ts|versions=6+
 file|tests/issue-14544/test.spec.ts|features=injectFn
 file|tests/issue-14544/service.spec.ts|versions=22
 file|tests/issue-14560/*.ts|features=injectFn
+file|tests/issue-14613/*.ts|versions=22
 file|tests/issue-9397/*.ts|features=injectFn
 file|tests/issue-10760/*.ts|features=injectFn
 file|examples/ngMocksGlobalMock/*.spec.ts|features=injectFn
--- a/tests/issue-14613/test.spec.ts
+++ b/tests/issue-14613/test.spec.ts
@@ -1,0 +1,24 @@
+import { effect, Injectable } from '@angular/core';
+
+import { MockBuilder, ngMocks } from 'ng-mocks';
+
+@Injectable({ providedIn: 'root' })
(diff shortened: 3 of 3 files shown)


satanTime:

PR #14619, opened 2026-08-11
fix(core): preserve Angular runtime services

satanTime:
## Why

The investigation for #14613 found the same root-provider mocking mechanism affecting
other Angular runtime primitives. `MockBuilder` can replace after-render and pending-task
infrastructure while constructing a kept root injectable.

That causes `afterNextRender()` callbacks to be discarded and allows
`ApplicationRef.whenStable()` to resolve before an unresolved `resource()` loader finishes.

## What

- Preserve `AfterRenderManager` and `AfterRenderImpl` by default.
- Preserve `PendingTasks` and `PendingTasksInternal` by default.
- Keep explicit `.mock(...)` overrides available.
- Add Angular 22 regressions for after-render callbacks and resource stability tracking.

## Impact

Kept root injectables retain Angular's after-render and stability contracts without requiring
`NG_MOCKS_ROOT_PROVIDERS` workarounds. Public application dependencies and unrelated defer
schedulers remain mockable.

Related to #14613.
Follow-up to #14614.

### DIFF
--- a/libs/ng-mocks/src/lib/common/core.config.ts
+++ b/libs/ng-mocks/src/lib/common/core.config.ts
@@ -22,6 +22,12 @@ export default {
     'DomSanitizer',
     'DomSanitizerImpl',
 
+    // Angular runtime primitives require their concrete root infrastructure.
+    'AfterRenderImpl',
+    'AfterRenderManager',
+    'PendingTasks',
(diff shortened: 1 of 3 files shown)


satanTime:

PR #14620, opened 2026-08-11
test(core): cover Angular 16+ runtime services

satanTime:
## Why

The investigation following #14613 found that the affected Angular runtime APIs predate
Angular 20: effects begin in Angular 16, `EffectScheduler` is used from Angular 17,
`afterNextRender()` begins in Angular 16, and `resource()` begins in Angular 19.

The fixes from #14614 and #14619 are version-independent, but their regressions did not cover
each API from its introduction. The older-version audit also confirmed that Angular 16's
`EffectManager` and `AfterRenderEventManager` do not need additional never-mock exceptions.

## What

- Run the #14613 effect regression on Angular 17+ with the available effect flush API.
- Add an Angular 16 render-cycle regression for its `EffectManager` predecessor.
- Run the after-render regression on Angular 16+ in zoned and zoneless profiles.
- Run the resource-stability regression on Angular 19+.
- Split the runtime tests and spread rules at their actual Angular API boundaries.
- Wait for fixture stability so after-render assertions work across both change-detection modes.

## Impact

The compatibility matrix now protects the runtime-service behavior from Angular 16 through
Angular 22 without adding speculative framework services to the never-mock policy. Explicit
mock overrides remain unchanged, and no documentation update is needed because public support
and behavior are unchanged.

Related to #14613.
Follow-up to #14614 and #14619.


satanTime:

PR #14654, opened 2026-08-17
fix(TestBed): respect explicit MockModule boundaries #14653

### LINKED ISSUE #14653: Bug: MockModule does not mock everything
### Description of the bug

`MockModule()` no longer mocks all directives declared in the given module. A directive with a host listener (`host: { '(toggled)': 'handleChangeEvent($event.target.checked)' }`) declared in the mocked module stays active: when the event is fired via `triggerEventHandler`, the original host listener is still invoked and throws:

TypeError: Cannot read properties of undefined (reading 'checked')
at _ToggleValueAccessor_toggled_HostBindingHandler

This worked with ng-mocks 14.15.3.

Link:
https://codesandbox.io/p/devbox/solitary-night-329jmz?file=%2Fsrc%2Ftest.spec.ts

**Expected:** All directives of a module imported via `MockModule(UiModule)` are mocked, so no original host listener is active. Only the component's own `(toggled)` handler runs and the test passes.

**Actual:** The original directive is not mocked and its host listener is executed. Because `triggerEventHandler` passes the given object directly as `$event`, `$event.target` is `undefined` and the listener throws `TypeError: Cannot read properties of undefined (reading 'checked')`, so the test fails.

satanTime:
## Why

The mixed TestBed bridge recursively preserves dependencies of real declarations and imports. That traversal continued through an explicitly supplied `MockModule`, so declarations inside the mocked module were added as inferred real dependencies. The original directive and its host listener therefore remained active.

## What

- Treat explicit mocks as dependency-traversal boundaries while rebuilding mixed TestBed configurations.
- Apply the same precedence to global mock, replace, and exclude resolutions while retaining explicit keeps.
- Keep shared dependencies real when another real TestBed path reaches them independently.
- Add cross-version regression coverage for a standalone host importing an explicitly mocked NgModule.

## Impact

`MockModule(UiModule)` once again mocks declarations beneath that module in mixed standalone TestBed configurations. Recursive preservation for unmocked real paths remains intact, including the behavior covered by #8649.

Fixes #14653

### DIFF
--- a/libs/ng-mocks/src/lib/common/func.extract-deps.ts
+++ b/libs/ng-mocks/src/lib/common/func.extract-deps.ts
@@ -11,6 +11,7 @@ export const funcExtractDeps = (
   result: Set<AnyDeclaration<any>>,
   recursive = false,
   visited = new Set<AnyDeclaration<any>>(),
+  shouldTraverse?: (dependency: AnyDeclaration<any>) => boolean,
 ): Set<AnyDeclaration<any>> => {
   const visit = (current: any): void => {
     if (visited.has(current)) {
@@ -35,7 +36,7 @@ export const funcExtractDeps = (
         // istanbul ignore if: it is here for standalone things, however they don't support modules with providers.
         const itemType = funcGetType(item);
         result.add(itemType);
-        if (recursive) {
+        if (recursive && (!shouldTraverse || shouldTraverse(itemType))) {
           visit(itemType);
         }
       }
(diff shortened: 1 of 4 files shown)


satanTime:
[v14.17.2](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.17.2) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14693, opened 2026-08-24
fix(MockRender): support readonly signal input bindings #14692

### LINKED ISSUE #14692: Bug: cannot assign signal input in DefaultRenderComponent
### Description of the bug

With v14.17.2, it does not work as expected (did not test other/earlier versions):

```ts
@Component()
export class MyComponent {
  myInput = input<string>(null);
}

MockRender(MyComponent).componentInstance.myInput = 'myValue';

// [ERROR] TS2540: Cannot assign to 'myInput' because it is a read-only property. [plugin angular-compiler]
```

After #14004 this should work, no?

### An example of the bug

see above

Signal inputs should be writable properties on the wrapper/container component.

satanTime:
Fixes #14692

Why:

- Angular recommends declaring signal input fields as `readonly`.
- `DefaultRenderComponent` currently carries that modifier onto its separate wrapper binding value.
- Tests therefore cannot assign a new signal input value through `MockRender`.

What:

- Add a regression for a styleguide-compliant readonly signal input.
- Remove `readonly` only from structurally detected signal-input binding keys.
- Preserve component modifiers for ordinary fields and signals, plus transformed-input write types.

Impact:

- `MockRender(...).componentInstance` exposes a writable host-side value for readonly signal inputs.
- The rendered component keeps its readonly `InputSignal` reference and Angular-managed update path.

Where:

- `tests/issue-14692/test.spec.ts`
- `test-spread.conf`
- `libs/ng-mocks/src/lib/mock-render/types.ts`

### DIFF
--- a/libs/ng-mocks/src/lib/mock-render/types.ts
+++ b/libs/ng-mocks/src/lib/mock-render/types.ts
@@ -47,6 +47,30 @@ type MockRenderInputBinding<T> = 0 extends 1 & T
       ? WriteT
       : T;
 
-export type DefaultRenderComponent<MComponent> = {
-  [K in keyof MComponent]: MockRenderInputBinding<MComponent[K]>;
+type MockRenderInputSignalKeys<MComponent> = {
+  [K in keyof MComponent]-?: 0 extends 1 & MComponent[K]
+    ? never
+    : [MockRenderInputSignalNode<MComponent[K]>] extends [never]
+      ? never
+      : K;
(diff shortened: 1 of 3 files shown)


satanTime:
[v14.17.3](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.17.3) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14775, opened 2026-09-02
fix(MockRender): hide internal query state from snapshots #14768

### LINKED ISSUE #14768: Bug: MockRender leaks internal __ngMocksPoint property into jest snapshots
### Description of the bug

After upgrading ng-mocks from 14.15.3 to 14.17.3 (Angular 22, jest-preset-angular 17 with its
`ng-snapshot` serializer), every `expect(fixture).toMatchSnapshot()` on a `MockRender` fixture
fails, because the `<mock-render>` wrapper now exposes an internal property that gets serialized:

```diff
- <mock-render>
+ <mock-render
+   __ngMocksPoint={[Function TargetComponent]}
+ >
```

The property appears to come from the signal-input support added in 14.16.x/14.17.x: when the
render has params, `__ngMocksPoint` is declared on the wrapper prototype as
`ViewChild(component, { static: true })` and is used by the input getters to unwrap signal inputs.
Because it is a plain enumerable property, snapshot serializers that print component properties
(jest-preset-angular's `ng-snapshot` prints all enumerable own keys of `componentInstance`) now
include it — breaking hundreds of existing snapshots with pure bookkeeping noise, including the
component class name, which additionally breaks snapshots on every component rename.

Notably, right next to it the wrapper defines `__ngContext__` with `enumerable: false`, which is
why that one never leaked. `__ngMocksPoint` just seems to be missing the same treatment.

### An example of the bug

```ts
import { Component, input } from '@angular/core';
import { MockRender } from 'ng-mocks';

@Component({
  selector: 'app-target',
  template: '<span>{{ value() }}</span>',
})
class TargetComponent {
  readonly value = input('default');
}

it('renders', () => {
  const fixture = MockRender(TargetComponent, { value: 'test' });
  expect(fixture).toMatchSnapshot();
});
```

With jest-preset-angular's `ng-snapshot` serializer configured, the stored snapshot now contains
`__ngMocksPoint={[Function TargetComponent]}` on the `<mock-render>` element.

Expected: internal ng-mocks bookkeeping properties do not show up in snapshots — upgrading
ng-mocks does not invalidate existing snapshots of unchanged components.

Actual: `__ngMocksPoint` is serialized into every MockRender snapshot, so all snapshot tests fail
after the upgrade (or, once updated with `-u`, bake internal state and class names into snapshots).

Suggested fix: define `__ngMocksPoint` (and `__ngMocks_index_*`, if enumerable) with
`enumerable: false`, the same way `__ngContext__` is defined in the same factory.
```

versions: ng-mocks 14.17.3, @angular/core 22.1.4, jest 30, jest-preset-angular 17.0.0

satanTime:
## Why

`MockRender` fixtures containing signal inputs expose the internal `__ngMocksPoint` query result as an enumerable property. The `jest-preset-angular` fixture serializer includes that property on `<mock-render>`, invalidating snapshots of otherwise unchanged components and tying them to component class names.

## What

- Define the query property as non-enumerable on each wrapper instance before Angular assigns its result, using the existing property helper.
- Add a Jest-only regression and expected fixture snapshot that check signal input rendering and the absence of internal bookkeeping in the serialized wrapper.

## Impact

Fixture snapshots no longer expose the internal query result. The property remains writable for Angular, preserving the signal input initialization and update path. Public documentation is unchanged because this restores the expected snapshot behavior.

Fixes #14768

### DIFF
--- a/e2e/jest/src/tests/issue-14768/__snapshots__/test.spec.ts.snap
+++ b/e2e/jest/src/tests/issue-14768/__snapshots__/test.spec.ts.snap
@@ -1,0 +1,11 @@
+// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+
+exports[`issue-14768 omits internal bookkeeping from fixture snapshots 1`] = `
+<mock-render>
+  <issue-14768-target>
+    <span>
+      test
+    </span>
+  </issue-14768-target>
+</mock-render>
+`;
--- a/e2e/jest/src/tests/issue-14768/test.spec.ts
+++ b/e2e/jest/src/tests/issue-14768/test.spec.ts
@@ -1,0 +1,26 @@
+import { Component, input } from '@angular/core';
+import { MockBuilder, MockRender } from 'ng-mocks';
+
+@Component({
+  selector: 'issue-14768-target',
+  standalone: true,
+  template: '<span>{{ value() }}</span>',
+})
+class TargetComponent {
+  public readonly value = input('default');
+}
+
+// @see https://github.com/help-me-mom/ng-mocks/issues/14768
+// The signal-input ViewChild becomes an enumerable wrapper property,
+// so the ng-snapshot serializer exposes internal state on <mock-render>.
+describe('issue-14768', () => {
+  beforeEach(() => MockBuilder(TargetComponent));
+
+  it('omits internal bookkeeping from fixture snapshots', () => {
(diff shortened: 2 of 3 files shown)


satanTime:
[v14.17.4](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.17.4) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14814, opened 2026-09-05
docs(router): complete guard coverage and documentation #1008

### LINKED ISSUE #1008: Cover all guards variants

satanTime:
## Why

#1008 asks whether all routing guard variants are covered, including the class-based `canLoad` guard from the original report. The implementation already handles all five guard properties, but core guard-filtering assertions covered only `canActivate`. The examples and guide also left the differences between cancellation, route matching, and lazy loading unclear.

## What

- Cover each guard property with class, function, and injection-token references, including nested routes, selective keeping and exclusion, provider overrides, directly supplied routes, and preservation of the original configuration.
- Add the original class-based `canLoad` scenario under `tests/issue-1008`, checking successful lazy loading after guard exclusion and suppression of the loader when a retained guard rejects navigation.
- Strengthen the existing functional examples with guards that fail if unexpectedly executed, navigation-result assertions, and lazy-loader assertions using the appropriate runner spy API. Correct the lazy module's component ownership and remove obsolete version checks already handled by the spread configuration.
- Explain each guard's navigation behavior, the functional API version boundary, guard/provider selection, standalone provider handling, and lazy-module testing in the guide and `MockBuilder` token reference. Align the snippets with the executable examples, include router root providers in the API setup, identify the Angular versions used by the simplified NgModule snippet, and describe the mocked service's initial state accurately.

## Impact

All five guard variants have explicit coverage and corresponding testing guidance. The existing runtime implementation already supports these cases, so this change completes tests and documentation without changing the published runtime behavior.

Closes #1008

### DIFF
--- a/docs/articles/api/MockBuilder.md
+++ b/docs/articles/api/MockBuilder.md
@@ -483,11 +483,16 @@ It covers `canActivate`, `canActivateChild`, `canDeactivate`, `canLoad`, and `ca
 with class guards, injection tokens, and functional guards.
 It's useful if you want to test a specific guard. 
 To do so, you need to [`.exclude`](#exclude) `NG_MOCKS_GUARDS` and to [`.keep`](#keep) the guard.
+Keep [`NG_MOCKS_ROOT_PROVIDERS`](#ng_mocks_root_providers-token) as well so the router's root services remain available.
 
 ```ts
 beforeEach(() => {
   return MockBuilder(
-      [RouterModule, RouterTestingModule.withRoutes([])],
+      [
+        RouterModule,
+        RouterTestingModule.withRoutes([]),
+        NG_MOCKS_ROOT_PROVIDERS,
+      ],
       ModuleWithRoutes,
     )
     .exclude(NG_MOCKS_GUARDS) // <- removes all guards
--- a/docs/articles/api/MockBuilder.md
+++ b/docs/articles/api/MockBuilder.md
@@ -478,7 +478,9 @@ even if it has been imported or declared in nested modules.
 
 ### `NG_MOCKS_GUARDS` token
 
-`NG_MOCKS_GUARDS` helps to **remove guards from all routes** in a test.
+`NG_MOCKS_GUARDS` helps to **remove guards from routes processed by `MockBuilder`**, including nested `children`.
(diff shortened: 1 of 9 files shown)


satanTime:

PR #14815, opened 2026-09-06
perf: reduce runtime overhead during specs

satanTime:
## Why

ng-mocks repeats dependency and prototype traversal, hook composition, and allocation during spec setup. Fixture searches also repeat DOM property reads and linear membership checks, while routine injector misses create exceptions. Builder promise entries can retain module metadata after resets. The build also repeats TypeScript runtime helpers across modules.

## What

- Visit shared dependency metadata once per resolution pass and use Sets for module, provider, and fixture traversal membership while preserving processing order.
- Discover methods and accessors in one prototype walk and avoid accessor replay allocations for ordinary mocked methods.
- Reuse hook composition and owned declaration registries, delay mock-map copies until needed, and keep the builder promise cache bounded and reset-aware.
- Use injector fallbacks for ordinary misses and reduce DOM getter work during instance searches.
- Add library performance specs for bounded DOM reads, shared cyclic traversal, registry reuse without input mutation, collection boundaries, and reset isolation.
- Use `ngMocks.ignoreOnConsole('warn')` in the zoneless input suite to suppress expected NG0914 notices in the core runner, with console methods restored after each spec.
- Bundle shared TypeScript helpers once per output and remove unused helpers in both production and readable development builds. Keep helper imports static and module scopes separate so direct eval does not prevent unused code removal.
- Replace custom collection-copy helpers with standard TypeScript syntax and native Map/Set copies. Iterate stable collections directly; retain snapshots where callbacks or dependency expansion can change the source, with an inline lint suppression explaining each case.

## Impact

Reduces runtime work during spec setup, mocking, instance searches, and teardown. Existing public APIs, accessor and spy behavior, provider priority, property-definition helpers, and supported Angular compatibility are preserved. Runtime and spec changes are under `libs/ng-mocks`; build configuration and a root-only `tslib` development dependency deduplicate the shipped helpers. The helpers are built into the distribution files, with no new public package dependencies. Existing assertions, performance thresholds, bundle-size limits, and ESLint configuration are unchanged. The new performance specs check work counts and allocation reuse without adding machine-dependent timing thresholds. The zoneless input suite suppresses warnings while the shared core runner retains Zone.js for its zoned tests. Public usage documentation does not need to change.


satanTime:

Issue #5465, opened 2023-04-19
Bug: `MockModule`, `NgFor` and `@happy-dom/jest-environment`


Yberion:
### Description of the bug

Heelo, it seems that when using `NgFor` instead of `CommonModule` (because I only use `*ngFor`), it does not work with Jest and `testEnvironment: '@happy-dom/jest-environment',`.

It works fine if I use `CommonModule`.

It does the following error:

```
TypeError: Class constructor CommonModule cannot be invoked without 'new'

      at new MockOfCommonModule (node_modules/ng-mocks/index.js:1:15515)
      at Object.MockOfCommonModule_Factory [as useFactory] (ng:\MockOfCommonModule\ɵfac.js:5:10)
      at Object.factory (node_modules/@angular/core/fesm2020/core.mjs:8118:38)
      at R3Injector.hydrate (node_modules/@angular/core/fesm2020/core.mjs:8031:35)
      at R3Injector.get (node_modules/@angular/core/fesm2020/core.mjs:7919:33)
      at R3Injector.<anonymous> (node_modules/ng-mocks/webpack:/ng-mocks/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts:371:53)
      at R3Injector.this (node_modules/ng-mocks/webpack:/ng-mocks/libs/ng-mocks/src/lib/mock-service/helper.create-clone.ts:10:65)
      at injectInjectorOnly (node_modules/@angular/core/fesm2020/core.mjs:633:33)
      at ɵɵinject (node_modules/@angular/core/fesm2020/core.mjs:637:42)
      at useValue (node_modules/@angular/core/fesm2020/core.mjs:7713:65)
      at R3Injector.resolveInjectorInitializers (node_modules/@angular/core/fesm2020/core.mjs:7968:17)
      at new NgModuleRef (node_modules/@angular/core/fesm2020/core.mjs:19817:26)
      at TestBedCompiler.finalize (node_modules/@angular/core/fesm2020/testing.mjs:23331:30)
      at TestBedImpl.get testModuleRef [as testModuleRef] (node_modules/@angular/core/fesm2020/testing.mjs:24294:49)
      at TestBedImpl.inject (node_modules/@angular/core/fesm2020/testing.mjs:24205:29)
      at TestBedImpl.createComponent (node_modules/@angular/core/fesm2020/testing.mjs:24257:44)
[...]

Issue #158, opened 2020-07-10
ES5 target support + Angular 10 Support


icorne:
Hello,

This is the only library I encountered thus far that does not support ES5 target anymore..
Could it be because in https://github.com/satanTime/ng-mocks/blob/83debd26ac89ed5ae5efb3c9be8ce7ad91ad6c64/e2e/a10es2015ivy/.browserslistrc you excluded IE11?

Can we try and revisit this?

satanTime:
Hi,

the lib does support es5, but angular 10 doesn't support it and always uses es2015 version of an imported lib.

I spent around 2 weeks to determine a solution and failed with every idea.

Feel free to try to fix it yourself, perhaps you'll find a working solution.

icorne:
I can have a look when I find some time. 
Where did you find that angular 10 does not support ES5 target?
We've got a pretty extensive app (hybrid with angularjs even) and everything is working except my tests which use ng-mocks.

satanTime:
During investigation, starting from angular 10 it ignores es5 build even it has been provided: https://github.com/ike18t/ng-mocks/blob/master/package.json#L5-L8.

I think in your app dependencies don't extend code base built by es5 and es2015, in case of ng-mocks if a component was built as es2015 and ng-mocks was built as es5 then it throws an error that an instance can't be created without using `new` operator that is required in es2015. Because es2015 is more popular it is the default build target.

Perhaps there's a way to downgrade to es5 not only the app itself, but also installed dependencies.

Current solution is to change tsconfig.spec.ts to use es2015.

satanTime:
Note: As an option to try: eval with detection of es6 classes.

satanTime:
Hi, might you check that https://github.com/ike18t/ng-mocks/pull/161 works for you on es5?

PR #14816, opened 2026-09-06
fix(core): removing eval

satanTime:
## Why

The ES5 package must extend native ES2015 Angular classes without calling their constructors as ordinary functions. The existing workaround evaluates a generated subclass. An eval failure falls back to ES5 inheritance and can produce `Class constructor CommonModule cannot be invoked without 'new'`, as seen with Happy DOM.

## What

Use `Reflect.construct` to invoke the parent constructor and preserve the instance prototype. A guarded construct trap preserves native `new.target` when consuming the ES5 bundle, and ordinary inheritance remains available when Reflect construction is absent. Preserve constructor stringification for Angular's reflection on Node 8, while retaining custom stringification methods.

Remove the evaluated declarations and temporary global properties. Add helper behavior specs for custom constructor stringification and construction when Proxy, Reflect, or Reflect.construct is unavailable.

## Impact

The constructor helper no longer requires eval. The change remains internal; public API documentation is unchanged.

Related to #158 and #5465.

### DIFF
--- a/libs/ng-mocks/src/lib/common/core.helpers.spec.ts
+++ b/libs/ng-mocks/src/lib/common/core.helpers.spec.ts
@@ -1,5 +1,91 @@
-import { extendClass, extractDependency } from './core.helpers';
+import {
+  extendClass,
+  extendClassicClass,
+  extractDependency,
+} from './core.helpers';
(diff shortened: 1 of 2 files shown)


satanTime:

PR #14853, opened 2026-09-09
fix(MockRender): preserve middleware signal inputs #14839

### LINKED ISSUE #14839: Bug:  MockRender destroys signal inputs on components rendered via the template middleware (selector-less or attribute selectors)
When `MockRender` renders a component whose selector is missing, or contains whitespace/commas/brackets, ng-mocks clones it into a `MockMiddleware<Name>` class and re-compiles it via JIT. That recompilation silently drops isSignal metadata, so every `input()` / `input.required()` is re-declared as an ordinary input. Angular then writes the bound value directly onto the instance field, overwriting the `InputSignal` function. The component's own template subsequently throws:

```
TypeError: ctx.myInput is not a function
    at templateFn (ng:///MockMiddlewareMyComponent.js:19:61)
```

**Reproduction**
```
@Component({
  // Any of these triggers it:
  //   selector omitted entirely
  //   selector: 'app-target[myInput]'
  //   selector: 'app-a, app-b'
  selector: 'app-target[myInput]',
  standalone: true,
  template: '{{ myInput() }}',
})
class TargetComponent {
  readonly myInput = input.required<string>();
}

describe('signal input + middleware', () => {
  beforeEach(() => MockBuilder(TargetComponent));

  it('fails', () => {
    MockRender(TargetComponent, { myInput: 'value' }); // TypeError: ctx.myInput is not a function
  });
});
```

Changing the selector to a plain app-target makes it pass.

**Expected**

`MockRender(Comp, { myInput: 'value' })` binds the raw value through the generated template into the real `InputSignal`, exactly as documented in `MockRende`r § `ComponentRef.setInput` and signal inputs. The selector should not affect this.

**Actual**

The `InputSignal` on the instance is replaced by the raw bound value, and any myInput() call in the template throws.

**Root cause**

In `libs/ng-mocks/src/lib/mock-render/func.reflect-template.ts`, the middleware is created whenever the selector is blank or "complex":

```
if (override.selector && /[\s,[\]]/.test(override.selector)) {
  override.selector = '';
}
if (!override.selector) {
  override.selector = `ng-mocks-${template.name}`;
  registerTemplateMiddleware(template, override);
}
```

`registerTemplateMiddleware` then re-decorates the clone by spreading the resolved metadata:
`Component({ changeDetection: ChangeDetectionStrategy.Default, ...meta, ...set })(child);`

`meta.inputs` is an array of { name, alias, required, `isSignal`, transform }. Angular's JIT parses that array with `parseInputsArray`, which hardcodes isSignal: false and has no branch for it:

```
results[value.name] = {
  bindingPropertyName: value.alias || value.name,
  classPropertyName: value.name,
  required: value.required || false,
  isSignal: false,                       // <-- metadata lost here
  transformFunction: value.transform != null ? new WrappedNodeExpr(value.transform) : null,
};
```

The only JIT path that honours `isSignal` is propMetadata:
```
inputsFromType[field] = {
  ...
  isSignal: !!ann.isSignal,
};
```

ng-mocks already uses that path correctly elsewhere — decorateInputs writes `__prop__metadata__` with isSignal, which is why mocked components keep callable signal inputs (cf. tests/issue-13671). The template middleware is the one place that routes signal inputs through the lossy array form.

Supporting evidence (behaviour matches the analysis exactly)
```
Call								 Result	Why
MockRender(Comp)					 throws	binds all inputs → every signal input overwritten
MockRender(Comp, {})				 passes	binds nothing → InputSignal survives untouched
MockRender(Comp, { x: signal(v) })	 passes	overwritten, but with a callable — passes by accident
MockRender(Comp, { x: v })			 throws overwritten with a non-callable
same component with a plain selector passes	no middleware, no recompile

```
**Suggested fix**

In registerTemplateMiddleware, do not hand signal inputs to Component() via the inputs array. Either apply them with the existing decorateInputs helper (prop-metadata path, which preserves isSignal), or split the list so only non-signal inputs go through meta.inputs:

satanTime:
## Why

`MockRender` recompiles components with missing or complex selectors under a generated selector. Angular JIT treats inputs declared only in the decorator's `inputs` array as ordinary inputs, losing the signal flag and replacing an `InputSignal` with its bound value.

## What

- Apply signal input property decorators to the middleware using the existing metadata helper, while preserving input overrides and ordinary input handling.
- Give the middleware its own input array so reflecting the clone cannot add duplicate bindings to the generated wrapper template.
- Add the issue reproducer for attribute, multiple, and missing selectors, covering required inputs, defaults, aliases, updates, transform counts, and ordinary inputs and outputs. Add source regressions for metadata isolation and override and alias precedence.

## Impact

Raw input bindings keep the rendered component's signal callable, and each update runs its transform once. This restores the documented `MockRender` contract, so public documentation and migration guidance do not change.

Fixes #14839

### DIFF
--- a/libs/ng-mocks/src/lib/mock-render/func.reflect-template.spec.ts
+++ b/libs/ng-mocks/src/lib/mock-render/func.reflect-template.spec.ts
@@ -1,0 +1,184 @@
+import {
+  Component,
+  Directive,
+  Input,
+  input,
+  reflectComponentType,
+} from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+
+import coreReflectDirectiveResolve from '../common/core.reflect.directive-resolve';
+
+import funcReflectTemplate from './func.reflect-template';
+
+describe('funcReflectTemplate', () => {
+  it('preserves signal inputs and ordinary input transforms on component clones', () => {
+    @Component({
+      standalone: false,
+      template: '',
+    })
+    class TargetComponent {
+      // Root tests transpile the fixture without Angular's signal transform.
+      @Input({ alias: 'publicValue', isSignal: true } as never)
+      public readonly value = input(0);
+
+      @Input({ alias: 'publicPlain', transform: String })
+      public plain = '';
+    }
+
+    const originalInputs =
+      reflectComponentType(TargetComponent)?.inputs;
+    const originalTemplateInputs = [
+      ...(coreReflectDirectiveResolve(TargetComponent).inputs || []),
+    ];
+    const configure = spyOn(
+      TestBed,
+      'configureTestingModule',
+    ).and.stub();
+
+    const template = funcReflectTemplate(TargetComponent);
+
+    const child =
+      configure.calls.mostRecent().args[0].declarations![0];
+    const inputs = reflectComponentType(child)?.inputs;
+    expect(inputs).toEqual(
+      jasmine.arrayContaining([
+        jasmine.objectContaining({
+          isSignal: true,
+          propName: 'value',
+          templateName: 'publicValue',
+        }),
+        jasmine.objectContaining({
+          isSignal: false,
+          propName: 'plain',
+          templateName: 'publicPlain',
+          transform: String,
+        }),
+      ]),
+    );
+    expect(inputs?.length).toEqual(2);
+    expect(reflectComponentType(TargetComponent)?.inputs).toEqual(
+      originalInputs,
+    );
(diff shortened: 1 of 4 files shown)


satanTime:
[v14.17.6](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.17.6) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14874, opened 2026-09-10
ci: share dependency caches across worker CPU models

satanTime:
## Why

PR #14860 failed in `jest:build` with `ng: not found` even though dependency installation and the Jest tests passed. CircleCI saved the dependencies under an `arch1-linux-amd64-6_85` cache key, but the build ran on a worker reporting `arch1-linux-amd64-6_106` and could not restore them.

The `{{ arch }}` template includes the CPU model, so identical Docker executors can produce different keys when CircleCI schedules jobs on different host hardware.

## What

Remove `{{ arch }}` from all dependency-cache restore and save keys. Keep the project prefixes, cache version, and package manifest and lockfile checksums. Give the Windows IE cache an explicit `a5es5-windows` prefix to keep its dependencies separate from Linux.

Document a commit-only exception for mounting shared Git metadata from linked worktrees, and align the issue workflow guidance with it. Hook setup uses temporary Git configuration; normal builds, tests, and pushes retain the existing isolated mounts.

## Impact

Install, test, and build jobs can share the same dependency cache across CircleCI's Linux amd64 CPU models. The new keys will populate fresh caches on their first use. Contributors can run normal commit hooks in linked worktrees while keeping working files isolated. Changes are limited to CI configuration and contributor guidance; public API documentation is unaffected.

Related to #14860.

### DIFF
--- a/.agents/skills/triage-issue/SKILL.md
+++ b/.agents/skills/triage-issue/SKILL.md
@@ -53,8 +53,8 @@ Create and maintain a plain Markdown checklist:
    - do not discard unrelated local changes in the original checkout; the worktree isolates the issue branch from them
    - run installs, builds, tests, checks, and commits from the issue worktree; do not use the original checkout as a
      fallback working directory or change its branch
-   - do not mount the original checkout or its `.git` directory into the worktree's Docker containers; report linked
-     Git metadata failures and discuss a supported solution without weakening worktree isolation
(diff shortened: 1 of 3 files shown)


satanTime:
[v14.17.6](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.17.6) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14893, opened 2026-09-10
build(docker): share one browser cache across worktrees

satanTime:
## Why

Each isolated Compose project currently creates its own Puppeteer cache, so worktrees download the same browser builds repeatedly. Legacy targets that support a configurable download directory also keep duplicate Chromium downloads in individual dependency directories.

## What

Share browser downloads through one external Docker volume named `ngmocks-puppeteer-cache`. Modern Puppeteer consumers use their native cache layout, and Angular 8–15 select revision-specific directories inside the same volume so legacy installer cleanup cannot remove another target's browser.

Create the volume from the existing install wrapper and document its lifecycle, legacy revision mappings, and the need to finish cold downloads before parallel installations. Align the lockfile-refresh workflow with those shared-cache requirements.

## Impact

Targets with configurable browser caches reuse downloads across services and worktrees. Browser installation remains enabled, without added skip flags or adapters. Angular 5–7 retain their existing installation because their Puppeteer versions lack a configurable download directory. Dependencies, Angular build caches, other Compose resources, and host browser installation retain their existing behavior. This changes development tooling and contributor guidance without changing the published library.


satanTime:

PR #14901, opened 2026-09-10
fix(MockRender): preserve aliased signal input bindings #14894

### LINKED ISSUE #14894: Investigate aliased signal input access in MockRender
## Background

Follow-up to [#14692](https://github.com/help-me-mom/ng-mocks/issues/14692) and its fix in [#14693](https://github.com/help-me-mom/ng-mocks/pull/14693) ([commit](https://github.com/help-me-mom/ng-mocks/commit/092dff9effbb03f698d32d6d4acd3379028d151e)). That fix makes signal input binding properties writable on `DefaultRenderComponent`, including inputs declared with `readonly`.

## Suspected gap

Source review suggests that a signal input with a public alias exposes an original-name wrapper property whose type and runtime behavior disagree. This remains an unverified scenario to reproduce before selecting a fix.

`DefaultRenderComponent` maps the original component property name to a writable binding value. The generated wrapper binds inputs by their public aliases, then mirrors remaining component properties. For an aliased signal input, this appears to leave the original-name property on the ordinary mirror path: its getter returns the signal, and its setter assigns directly to the real component property. A scalar assignment accepted by the wrapper type could therefore replace the signal reference.

Relevant paths: [binding-value types](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/libs/ng-mocks/src/lib/mock-render/types.ts#L50-L64), [public-alias bindings](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/libs/ng-mocks/src/lib/mock-render/func.create-wrapper.ts#L159-L175), and [mirror setter](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/libs/ng-mocks/src/lib/mock-render/func.install-prop-reader.ts#L35-L44).

This does not assert a failure of the documented public-alias workflow, `fixture.componentRef.setInput('publicValue', value)`, or explicit params using `publicValue`.

## Minimal scenario

In a compiled Angular target supporting signal inputs, configure `MockBuilder(Target)` and exercise a render without params:

```ts
@Component({
  selector: 'target',
  standalone: true,
  template: '{{ value() }}',
})
class Target {
  readonly value = input('initial', { alias: 'publicValue' });
}

const fixture = MockRender(Target);
const originalSignal = fixture.point.componentInstance.value;

fixture.componentInstance.value = 'updated';
fixture.detectChanges();

expect(fixture.point.componentInstance.value).toBe(originalSignal);
expect(originalSignal()).toBe('updated');
expect(fixture.nativeElement.textContent).toContain('updated');
```

The assignment is the operation advertised by the current original-name wrapper type; the expected assertions describe safe support for that operation.

## Expected outcome

The wrapper's type and runtime behavior should agree. If original-name binding access is supported, reads should expose the binding value and writes should update the public binding without replacing the real signal. If only alias-based access is intended, the exposed type and documentation should make that boundary explicit instead of advertising scalar access through the original name.

- Establish the documented public-alias `setInput` and explicit-params workflows as controls, then reproduce original-name reads and writes with `MockRender(Target)`.
- Cover the chosen contract in a compiled signal-input regression and the relevant type coverage. Preserve the real signal's identity and rendered updates for supported writes.
- Preserve unaliased readonly signal input assignment from #14692, accepted transform write types, and readonly protection for ordinary non-input properties.
- Compare with the existing [ordinary aliased input mirror case](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/tests/mock-render-mirrors-component/test.spec.ts#L98-L107) and [public-alias signal case](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/tests/issue-11101/test.spec.ts#L63-L99). Keep any source change focused on the demonstrated mismatch.

satanTime:
## Why

`MockRender(Component)` exposes signal inputs as writable binding values on its default fixture type. For an aliased input, the generated wrapper used the public alias while the original property name fell through to the ordinary component mirror. Reading that property returned a signal function, and assigning a binding value could replace the rendered component's signal.

Follow-up to #14692 and #14693. Fixes #14894.

## What

Forward original signal-input property names to their existing public wrapper bindings when rendering without explicit params. Preserve names already owned by the wrapper, including other public input aliases.

Add regressions for readonly aliased inputs and models, transformed inputs and raw binding values, signal identity, rendered updates, model output behavior, public-alias params and `setInput`, alias collisions, and ordinary input mirrors.

## Impact

Original-name fixture reads and writes now agree with the advertised binding-value type without replacing the target's signal. Explicit params and binding lists retain their current behavior. The fix restores the documented fixture contract, so no public documentation change is required.


satanTime:

PR #14902, opened 2026-09-11
test(TestBed): cover nested mock module boundaries #14895

### LINKED ISSUE #14895: Cover nested MockModule boundaries and shared dependency precedence
## Why

Follow-up to [#14653](https://github.com/help-me-mom/ng-mocks/issues/14653), resolved by [#14654](https://github.com/help-me-mom/ng-mocks/pull/14654) ([fix commit](https://github.com/help-me-mom/ng-mocks/commit/8c22ef641541b497c5394a747d1fa4590362b9cf)).

The fix prevents the mixed real/mock TestBed bridge from inferring that dependencies below an explicitly mocked module must remain real. The reported host-listener failure has a regression, but the implementation also defines behavior for nested modules, shared dependencies and global resolutions. Those combinations are not directly covered by the issue regression.

This is a coverage follow-up, not a confirmed remaining runtime defect. The existing shared-component tests in `tests/issue-8649/test.spec.ts` already cover explicit component mocks, shared real imports and a global mock. The global module tests exercise replacement, exclusion and keeping outside the mixed TestBed bridge.

## Expected behavior

- Dependencies reachable only through an explicit `MockModule` boundary remain mocked, including nested module declarations and providers. Their real constructors, host listeners and provider factories do not run.
- A shared dependency remains real when another real import path independently reaches it. Reordering the top-level TestBed imports does not change that result.
- Explicit TestBed selections and global keep/mock/replace/exclude resolutions retain their existing precedence. Inferred keeps do not widen the requested real dependency graph across a mock, replacement or exclusion boundary.

- [ ] Extend the relevant feature coverage with a real host and an explicitly mocked module containing a nested module. Check mock identities and the absence of real declaration/provider side effects, alongside the host's expected behavior.
- [ ] Add a shared descendant reachable through both that mocked boundary and an independent real path. Check that the shared descendant stays real, private descendants stay mocked, and reversing the TestBed root order preserves the result.
- [ ] Cover global keep, replace and exclude at a module boundary in mixed real/mock TestBed configuration, checking the selected declarations and their descendants.
- [ ] Add a focused `funcExtractDeps` unit case for `shouldTraverse`: collect the blocked node without its descendants, and allow a later independent real root to reach a shared descendant.
- [ ] Preserve the existing #14653 and #8649 regressions. Change source behavior only if an added regression demonstrates a failure, and keep any resulting fix limited to the failed boundary or precedence rule.

The relevant implementation is the mixed TestBed traversal in `libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts` and the traversal predicate in `libs/ng-mocks/src/lib/common/func.extract-deps.ts`.

satanTime:
## Why

The mixed TestBed fix in #14654 stopped inferred real dependencies at explicitly mocked module boundaries. Its original host-listener regression did not directly cover nested modules, shared descendants reached through an independent real import, or global module resolutions.

Follow-up to #14653 and #14654. Fixes #14895.

## What

Add real-Angular regressions for nested mocked declarations and providers, preserved real host behavior, shared real descendants, and reversed TestBed import order. Check declaration identities and the absence of real constructor, listener, and provider-factory side effects below mock boundaries.

Extend the global module feature suites with mixed-TestBed cases for mock, keep, replace, and exclude, including a globally mocked child module inside a globally kept parent. Add focused dependency-traversal cases showing that blocked nodes are collected without visiting their descendants and that independent real roots still reach shared descendants in either root order.

## Impact

The broader boundary and precedence contracts now have regression coverage alongside the original issue tests. Existing behavior is preserved, so no public API or documentation change is required.

### DIFF
--- a/libs/ng-mocks/src/lib/common/func.extract-deps.spec.ts
+++ b/libs/ng-mocks/src/lib/common/func.extract-deps.spec.ts
@@ -21,6 +21,11 @@ class Issue7490Level2Module {}
 })
 class Issue7490Level1Module {}
 
+@NgModule({
+  imports: [Issue7490Level3Module],
+})
+class Issue14895RealRootModule {}
+
 describe('funcExtractDeps', () => {
   it('collects recursive dependencies to their full depth', () => {
     const result = funcExtractDeps(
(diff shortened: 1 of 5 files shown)


satanTime:

PR #14903, opened 2026-09-11
fix(core): preserve Angular effect and render managers #14896

### LINKED ISSUE #14896: Preserve Angular 16 runtime managers during directly kept service construction
## Context

Follow-up to [#14613](https://github.com/help-me-mom/ng-mocks/issues/14613), fixed by [#14614](https://github.com/help-me-mom/ng-mocks/pull/14614), with adjacent runtime infrastructure covered by [#14619](https://github.com/help-me-mom/ng-mocks/pull/14619) and additional compatibility tests in [#14620](https://github.com/help-me-mom/ng-mocks/pull/14620).

The original report concerned a mocked Angular effect scheduler during construction of a directly kept service. Angular 16 uses the earlier root services `EffectManager` and `AfterRenderEventManager`. The current preservation list contains their newer counterparts, but neither earlier name nor its underscore variant.

This is a source-supported risk; the scenario below has not yet been reproduced at runtime.

## Source evidence

- [The runtime infrastructure preservation list](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/libs/ng-mocks/src/lib/common/core.config.ts#L25) includes `EffectScheduler`, `AfterRenderManager`, and `AfterRenderImpl`.
- [Runtime injection can mock an unrecognized root class](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/libs/ng-mocks/src/lib/common/ng-mocks-runtime-inject.ts#L19) while a kept service is constructed.
- The existing [Angular 16 effect regression](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/tests/provider-never-mock/angular-16-effect-manager.spec.ts#L28) and [after-render regression](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/tests/provider-never-mock/angular-16-plus-after-render.spec.ts#L32) use `MockBuilder(HostComponent).keep(TargetService)`.
- That chained `.keep()` does not select the service wrapper's [`shallow === false` branch](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/libs/ng-mocks/src/lib/mock-builder/promise/create-runtime-inject-provider.ts#L36). These fixtures therefore do not exercise the directly kept service path used by the original report.

## Minimal scenario

On Angular 16, enable the runner's auto-spy mode and use the existing effect fixture with the service included in the first argument:

```ts
beforeEach(() => MockBuilder([EffectTarget, EffectHostComponent]));

it('runs the effect of a directly kept service', async () => {
  const fixture = MockRender(EffectHostComponent);
  await fixture.whenStable();

  expect(fixture.point.componentInstance.target.called).toBe(true);
});
```

`EffectTarget` is a root injectable whose constructor registers an `effect()` that sets `called = true`; the host injects that service. Repeat the same setup for the existing service whose constructor calls `afterNextRender()`.

The service needs to be directly kept so construction enters the runtime injection wrapper. The host supplies the render cycle required by Angular 16.

## Acceptance criteria

- Add real Angular 16 regressions for both `effect()` and `afterNextRender()` during directly kept service construction with auto-spy enabled.
- Confirm the regressions exercise the service wrapper and reproduce any missing-manager failure before changing the preservation policy.
- If reproduced, preserve the required Angular 16 infrastructure through the existing policy, including supported compiled name variants.
- Keep explicit mocking decisions effective and ordinary application root services mockable.
- Retain the existing Angular 17+ effect and newer after-render/resource coverage.

Keep this follow-up scoped to the earlier runtime managers and the missing service-construction coverage.

satanTime:
## Why

The EffectScheduler fix in #14614 and the runtime-service preservation in #14619 omitted the earlier `EffectManager` and `AfterRenderEventManager` classes. With auto-spy enabled, constructing a directly kept service could replace those framework services with empty mocks. Angular 16 effects then never ran, and after-render callbacks registered by services in Angular 16 and 17 never fired.

The earlier compatibility fixtures used chained `.keep(Service)`, which did not enter the directly kept service's runtime-injection wrapper.

Fixes #14896. Follow-up to #14613, #14614, #14619, and #14620.

## What

Preserve both earlier manager names and their underscore variants through the existing shared framework-provider policy. Let an explicit mock resolution override that default during runtime injection, matching the existing provider policy. This also covers globally mocked root classes that have runtime provider definitions without decorator metadata.

Add direct-service regressions for effect execution and signal updates, one-time after-render callbacks, and continued mocking of application root dependencies without real constructor or method side effects. Add focused source cases for all name variants, configured injector results, and explicit mock precedence. Add real-Angular cases for global and explicit mocks of metadata-only root classes, including configured values, injection identity, and absence of real constructor side effects.

## Impact

Services using the earlier Angular managers retain working framework callbacks while their application dependencies remain mockable. Explicit provider choices and the existing newer scheduler and render-manager behavior remain intact. This restores the existing contract, so no public API or documentation change is required.


satanTime:

PR #14584, opened 2026-08-09
docs(migrations): explain signal input changes #14565

satanTime:
## Why

ng-mocks 14.16.0 corrected mocked signal inputs so they retain the same callable interface as real components. Tests that relied on the previous scalar behavior can therefore need updates, as reported in #14565.

## What

- Add an ng-mocks 14.15 to 14.16 migration section for signal inputs.
- Show how to update direct mock-instance assertions and how to use `ngMocks.input` across both versions.
- Explain how to prepare for the update, how to change input values safely and which tests are unaffected.

## Impact

Angular 17+ users can identify the affected tests before upgrading and migrate them without changing application components or decorator-based inputs.

Closes #14565

Issue #14565, opened 2026-08-07
Update to version 14.16.0 had breaking changes


ShaneWhittet:
If possible, add an upgrade guide/script, or hold changes to a major version.

Any suggestions appreciated for this version change. 

Thank you!

satanTime:
There are no breaking changes. If you encounter a bug, please use a bug template to report it.

GipHub123:
@satanTime I believe that @ShaneWhittet was referring to this issue https://github.com/help-me-mom/ng-mocks/issues/13671.
I had to fix/change most of my test cases, although the changes were relatively easy to make

satanTime:
[v14.17.0](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.17.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

Issue #13671, opened 2026-04-24
Bug: signal inputs of mocked component exposed as non signal fields


Tukkan:
### Description of the bug
When mocking a component with signal inputs, the input field are exposed as non signal fields, which resuilts with accessing the field without braces when mocked and with braces when used with `keep`

According to docummentation [(https://ng-mocks.sudo.eu/api/MockComponent)](https://ng-mocks.sudo.eu/api/MockComponent), 
`The class of a mock component has: the same @Inputs and @Outputs with alias support`
It is indeed mentioning only old syntax with @Input. The new input() syntax do not apply to this rule.

When accessing a signal input of mocked component with braces, eg. name(), it results with an error:
`TypeError: de.componentInstance.name is not a function`

### An example of the bug
```typescript
import { Component, input } from '@angular/core';
import { MockBuilder, MockRender, ngMocks } from 'ng-mocks';

@Component({
  selector: 'app-name1',
  template: `{{ name() }}`,
})
export class Name1Component {
  readonly name = input('');
}

@Component({
  selector: 'app-name2',
  template: `{{ name() }}`,
})
export class Name2Component {
  readonly name = input('');
}

@Component({
  template: `
    <app-name1 [name]="'hello1'"></app-name1>
    <app-name2 [name]="'hello2'"></app-name2>
  `,
  imports: [Name1Component, Name2Component],
})
class WrapperComponent {}

describe('Given WrapperComponent', () => {
  beforeEach(
    () => MockBuilder(WrapperComponent)
      .keep(Name1Component)
      .mock(Name2Component), // reduntant, mocked by MockBuilder anyways, added for clarity
  );

  it('should set proper names', () => {
    MockRender(WrapperComponent);

    const instance1: Name1Component = ngMocks.find(Name1Component).componentInstance;
    const instance2: Name2Component = ngMocks.find(Name2Component).componentInstance;

    expect(instance1.name()).toBe('hello1'); // used with keep, works
    expect(instance2.name()).toBe('hello2'); // used with mock, throws instance2.name is not a function
    expect(instance2.name).toBe('hello2');   // works
  });
});
```

satanTime:
[v14.16.0](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.16.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14904, opened 2026-09-11
docs(migrations): clarify mocked signal input behavior #14897

### LINKED ISSUE #14897: Document and cover signal inputs on mocked directives
## Why

Follow-up to [#14565](https://github.com/help-me-mom/ng-mocks/issues/14565), resolved through migration documentation in [#14584](https://github.com/help-me-mom/ng-mocks/pull/14584) ([documentation commit](https://github.com/help-me-mom/ng-mocks/commit/74f288a761c0821c1f8eefc5810aae6596c1fd6b)). The underlying signal-input behavior change was introduced by [#14490](https://github.com/help-me-mom/ng-mocks/pull/14490) for [#13671](https://github.com/help-me-mom/ng-mocks/issues/13671).

The ng-mocks 14.15-to-14.16 migration section describes callable signal inputs on mocked components. Signal-input initialization is implemented in the shared `Mock` base, which is also used by `MockDirective`. The guide therefore omits another affected declaration type.

Existing regressions cover mocked component signal inputs and shared-base initialization. Broader metadata tests cover real directive metadata, but do not directly exercise compiled `MockDirective` signal-input initialization and binding. This follow-up concerns that specific documentation and regression gap; it does not assert a confirmed directive runtime defect.

## Expected behavior

- Bound `input()` and `input.required()` properties on a mocked directive remain callable Angular signals, with aliases, required metadata and transforms preserved.
- Updating the parent binding updates the signal value, and `ngMocks.input()` returns that value.
- Direct `MockDirective` use and directives mocked through `MockBuilder` expose the same signal-input interface.
- Kept directives continue to expose their real signal inputs, and decorator-based `@Input()` properties retain their ordinary property behavior.
- Migration guidance explains direct signal reads and binding updates for components and directives, including directives mocked indirectly through modules. It does not imply that mocked declarations preserve original field initializers or constructor behavior.

- [ ] Add a compiled attribute-directive regression with ordinary, aliased, required and transformed signal inputs. Exercise both direct `MockDirective` and `MockBuilder` mocking.
- [ ] Change parent bindings twice and assert signal identity/callability, values, transform results and `ngMocks.input()` lookup. Include a kept directive and a decorator input as preservation cases.
- [ ] Gate the regression at the actual signal-input API boundary and follow the repository's compiler-metadata guard pattern where needed.
- [ ] Update the existing migration section to cover mocked components and directives, using the regression's supported read/update behavior and accurately identifying unaffected cases.
- [ ] Change source behavior only if the added compiled regression demonstrates a failure. Keep any resulting fix limited to the failing directive signal-input path.

The relevant implementation is `applyInputs` in `libs/ng-mocks/src/lib/common/mock.ts`, reached by `DirectiveMockBase` in `libs/ng-mocks/src/lib/mock-directive/mock-directive.ts`. The migration text is in `docs/articles/migrations.md`; existing component coverage includes `tests/issue-13671/test.spec.ts`.

## Clarified transform scope

The initial expectation of automatic signal initializer-transform preservation needs correction. Angular omits these callbacks from reflected declaration metadata and stores them on the signal created by the original initializer. Recovering arbitrary callbacks would require running original initialization, which mocks intentionally skip.

The completion criterion for this follow-up is therefore to document and cover the existing distinction: mocked component and directive signals receive raw bound values without original transform side effects; kept declarations retain their real signal transforms. Reflected decorator-input transforms remain preserved. The migration guide and the linked MockComponent API article should explain this distinction, including `.keep()` for tests that need the real transform logic.

satanTime:
## Why

The 14.15-to-14.16 migration guide describes the signal-input change only for component mocks, although the same initialization code is used by directive mocks, including directives mocked through modules.

The migration guide and MockComponent API article also promise preservation of signal-input transforms. Angular does not include transforms declared inside `input()` initializers in reflected input metadata. Since mocks skip original initialization, they receive raw binding values instead: a bound string `'2'` stays a string even when the original declaration transforms it into a number.

Fixes #14897. Follow-up to #14565, #14584, and #13671.

## What

Extend the migration guidance to components and directives, direct mocks and module-based mocking, callable reads, public aliases, and host-binding updates. Clarify that `ComponentRef.setInput` applies to components and that kept declarations are the path for testing original signal transforms or initialization.

Add compiled regressions for direct MockDirective use, MockBuilder module mocking, a kept directive, and direct MockComponent use. Cover ordinary, aliased and required signal inputs, stable signal identity across repeated updates, value lookup, ordinary decorator inputs, and absence of original initialization or transform side effects on mocks. The kept control retains its real transform behavior. Extend the signal-initialization unit coverage to check missing required values, and document change notifications for plain TestBed hosts in zoneless tests.

Correct both documentation pages to distinguish signal initializer transforms from reflected decorator-input transforms. Preserve the existing migration heading anchor.

## Impact

The guides now describe the behavior users can rely on when upgrading and show how to retain real transform logic where needed. The change updates documentation and regression coverage without altering runtime behavior or public APIs.

### DIFF
--- a/docs/articles/api/MockComponent.md
+++ b/docs/articles/api/MockComponent.md
@@ -26,7 +26,7 @@ The class of a mock component has:
 
 - the same `selector`
 - the same inputs and outputs with alias support
-- signal inputs which stay callable and keep their required and transform metadata
+- signal inputs which stay callable when declared with `input()` or `input.required()`
 - templates with pure `<ng-content>` tags to allow transclusion
 - support for `@ContentChild` and `@ContentChildren`
 - support for `ControlValueAccessor`, `Validator` and `AsyncValidator`
@@ -101,8 +101,8 @@ describe('Test', () => {
 
 Starting with Angular 17, inputs can be declared with `input()` and
 `input.required()`. `MockComponent` preserves them as signals instead of
-replacing them with ordinary properties. Angular can therefore bind aliases
-and run input transforms in the same way as it does for the real component.
+replacing them with ordinary properties. Angular updates their values through
+template bindings, including bindings that use input aliases.
 
 ```ts
 @Component({
@@ -134,9 +134,15 @@ expect(child.name()).toEqual(fixture.point.componentInstance.name);
 ```
 
 The template uses the input alias, while the mock instance is read through the
-original property name. Required signal inputs remain required in Angular's
-metadata. For details about binding signal inputs on the component rendered by
+original property name. For details about binding signal inputs on the component rendered by
 `MockRender`, see [ComponentRef.setInput and signal inputs](MockRender.md#componentrefsetinput-and-signal-inputs).
+
+Mocks do not run the original component's constructor or field initializers.
+Angular stores transforms supplied to `input()` or `input.required()` on the initialized signal,
+without exposing them in reflected input metadata. A mock therefore receives the raw bound value
+without running that transform; for example, a bound string `'2'` remains a string.
+Decorator-based `@Input({ transform: ... })` transforms are preserved through their reflected metadata.
+Use `.keep(ChildComponent)` when the test needs the real component's initialization or signal-input transform.
 
 ## Standalone components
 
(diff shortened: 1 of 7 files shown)


satanTime:

PR #14905, opened 2026-09-11
fix(MockRender): schedule zoneless wrapper and template updates #14898

### LINKED ISSUE #14898: Investigate zoneless scheduling for custom templates and wrapper writes
## Background

Follow-up to [#12061](https://github.com/help-me-mom/ng-mocks/issues/12061) and its fix in [#14563](https://github.com/help-me-mom/ng-mocks/pull/14563) ([commit](https://github.com/help-me-mom/ng-mocks/commit/40714906cf07723f2de2e2a187d797b928be0e84)). The fix adds scheduling for params updates in zoneless `MockRender` and `MockRenderFactory` usage.

## Suspected gaps

Source review suggests two adjacent update paths may bypass that scheduling. Both remain unverified scenarios to reproduce and compare with the intended API contract before selecting a fix.

1. **Custom-template params.** String templates use empty declaration metadata, so the generated wrapper has no `inputBindings` for the scheduler to instrument. A params property used by `<target [value]="value">` can therefore change without notifying Angular, even though class-based rendering instruments the corresponding input. See [string-template metadata](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/libs/ng-mocks/src/lib/mock-render/mock-render-factory.ts#L401) and [scheduler installation](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/libs/ng-mocks/src/lib/mock-render/mock-render-factory.ts#L275-L280).
2. **Wrapper proxy writes.** In class-based rendering with params, direct params assignment uses the scheduling setter. Assignment through `fixture.componentInstance` instead uses the saved raw writer, apparently bypassing `markForCheck`. The docs describe these two ways of changing params as equivalent. See [scheduling setter and raw writer](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/libs/ng-mocks/src/lib/mock-render/mock-render-factory.ts#L109-L145), [proxy write path](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/libs/ng-mocks/src/lib/mock-render/func.install-prop-reader.ts#L35-L44), and [proxy contract](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/docs/articles/api/MockRender.md#L37-L49).

The suspected behavior concerns automatic scheduling followed by `whenStable()`. It is not a claim that an explicit `fixture.detectChanges()` is broken.

## Minimal scenarios

In a zoneless TestBed, configure `MockBuilder(Target)` with this target:

```ts
@Component({
  selector: 'target',
  standalone: true,
  changeDetection: ChangeDetectionStrategy.OnPush,
  template: '{{ value }}',
})
class Target {
  @Input() value = 'initial';
}
```

Custom-template params update:

```ts
const props = { value: 'initial' };
const fixture = MockRender('<target [value]="value"></target>', props);
await fixture.whenStable();

props.value = 'updated';
await fixture.whenStable();

expect(ngMocks.findInstance(Target).value).toBe('updated');
expect(fixture.nativeElement.textContent).toContain('updated');
```

Separately, a class-based wrapper proxy update:

```ts
const props = { value: 'initial' };
const fixture = MockRender(Target, props);
await fixture.whenStable();

fixture.componentInstance.value = 'updated';
await fixture.whenStable();

expect(props.value).toBe('updated');
expect(fixture.point.componentInstance.value).toBe('updated');
expect(fixture.nativeElement.textContent).toContain('updated');
```

## Expected outcome

Supported mutations of wrapper bindings should schedule the update needed for a subsequent `whenStable()` to expose the new input and rendered value. Establish whether this contract includes custom-template params and wrapper proxy writes. If either deliberately requires manual change detection, document that distinction explicitly.

- Settle each fixture before mutation so initial rendering cannot conceal a missing notification. Compare both scenarios with direct class-based params writes and `ComponentRef.setInput` controls.
- Cover custom templates through both `MockRender` and `MockRenderFactory(template, ['value'])`, with explicit `detectChanges()` as a separate rendering control.
- Cover wrapper proxy writes with params, including a subsequent direct params write, so both routes continue to update the same value. The [existing regression](https://github.com/help-me-mom/ng-mocks/blob/2e545369778b4a8cd1c78fe030abb3f1e3610d2e/tests/issue-12061/test.spec.ts#L108-L144) uses manual change detection immediately after the proxy write.
- Preserve class-based ordinary and signal input updates, OnPush behavior, accessor semantics, and the existing handling of params descriptors that cannot be instrumented. Retain zoned manual-change-detection behavior.
- Keep any fix focused on the reproduced scheduling gap. Decide and document the custom-template params policy before extending notification to unrelated template expressions.

satanTime:
## Why

The zoneless scheduling fix for #12061 instruments direct class-based params assignments, but wrapper proxy writes bypass that notification. Custom templates have no reflected input list, and renders without params own their input setters on the wrapper instead. After a fixture settles, these updates can leave its inputs and rendered content stale despite awaiting `whenStable()`.

Fixes #14898. Follow-up to #12061 and #14563.

## What

Route wrapper writes through the params object's current setter so all live fixtures sharing those params receive their notifications. Instrument the wrapper's own inputs when params are absent, select the declared binding context for custom templates, and avoid notifying destroyed fixture views while preserving the original setter behavior.

Add regressions for class and custom-template rendering, MockRenderFactory, params and wrapper writes, ordinary and signal inputs, signal initialization and transforms, inherited accessors, shared params, and fixture destruction. Cover the Angular 19 experimental provider separately from the stable zoneless API. Retain explicit change detection and ComponentRef.setInput controls, and cover custom-template key selection, callback binding, and nested mutations.

Document the zoneless update contract and the limits of automatic scheduling in the MockRender API article.

## Impact

Supported top-level binding writes become visible after `whenStable()` across the documented params and wrapper interfaces. Existing descriptor restrictions, callback receivers, signal identity and initialization, and explicit change-detection behavior remain intact. Custom templates observe their selected binding keys without parsing arbitrary template expressions or observing deep mutations.


satanTime:

PR #14906, opened 2026-09-11
fix(MockBuilder): preserve runtime dependency policies #14899

### LINKED ISSUE #14899: Cover inject() parity across classic declarations and default auto-spy mode
## Context

Follow-up to [#14560](https://github.com/help-me-mom/ng-mocks/issues/14560), fixed by [#14564](https://github.com/help-me-mom/ng-mocks/pull/14564). The shared runtime injection mechanism originated in [#14508](https://github.com/help-me-mom/ng-mocks/pull/14508) for [#9397](https://github.com/help-me-mom/ng-mocks/issues/9397), and [#14557](https://github.com/help-me-mom/ng-mocks/pull/14557) extended it to kept services for [#14544](https://github.com/help-me-mom/ng-mocks/issues/14544).

The classic component fix also selects directives and pipes, but its regression suite covers components only. Its tests, and the related kept-service tests, all enable auto-spy. We should cover the other declaration kinds and settle whether switching off auto-spy should change which dependencies are mocked.

The default-mode behavior is a source-supported policy gap, not a reproduced runtime failure. The directive and pipe cases are missing coverage; no defect in those paths is established.

## Source evidence

- [The declaration selection branch](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/libs/ng-mocks/src/lib/mock-builder/promise/create-runtime-inject-provider.ts#L32) handles components, directives, and pipes.
- [The issue regression](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/tests/issue-14560/test.spec.ts) covers classic component field initializers and constructor-body `inject()` calls with auto-spy enabled.
- [Runtime mocking requires a custom spy factory](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/libs/ng-mocks/src/lib/common/ng-mocks-runtime-inject.ts#L22), so `ngMocks.autoSpy('default')` bypasses that mechanism.
- [Auto-spy documentation](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/docs/articles/extra/auto-spy.md#L89) describes default mode as using empty methods instead of spies. [MockBuilder documentation](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/docs/articles/api/MockBuilder.md#L568) describes root mocking without an auto-spy prerequisite.
- The original [#14508 description](https://github.com/help-me-mom/ng-mocks/pull/14508) refers to auto-spied root services but does not explicitly document a limitation for default mode.

## Minimal scenarios

For policy parity, use a root dependency whose constructor increments a counter and whose `echo()` returns a real value. Compare two directly kept classic components: one receives it through a constructor parameter, and the other assigns `dependency = inject(TargetDependency)`.

Set `ngMocks.autoSpy('default')`, then use one-argument `MockBuilder(TargetComponent)` and render each component. If default mode follows the documented mocking policy, both dependencies should have empty `echo()` methods returning `undefined`, and neither should execute the real dependency constructor. No spy matcher is needed. Include the equivalent directly kept service case because it shares the same guard.

For declaration coverage, repeat the existing component behavior with a kept non-standalone directive and pipe. Resolve the dependency in field initializers and constructor bodies, and assert the mock is already available during initial construction or rendering.

## Acceptance criteria

- Add focused real-Angular regressions for non-standalone directives and pipes under the supported one-argument MockBuilder mode.
- Add constructor-injection versus `inject()` parity cases with auto-spy disabled for a classic declaration and a directly kept service.
- Establish the intended default-mode policy. If default mode should still mock dependencies, fix the shared runtime guard after demonstrating the divergence. If the dependency-policy distinction is intentional, document it explicitly and test both modes.
- Preserve explicit keep/mock/exclude choices, kept-module behavior, and the existing two-argument MockBuilder policy.
- Use the actual API compatibility boundaries for the regression files and retain the existing component and standalone declaration coverage.

Keep this follow-up scoped to declaration coverage and auto-spy policy parity. Injector lookup flags and Angular runtime-manager preservation are separate follow-ups.

satanTime:
## Why

Runtime `inject()` mocking depended on a custom auto-spy factory. Disabling auto-spy therefore constructed real root dependencies in kept components, directives, pipes, and services even when constructor injection followed the expected mocking policy. Explicit per-builder keep and exclude decisions were also lost when the temporary builder state was restored, including root dependencies left to Angular's fallback in two-argument builder setups.

Follow-up to #14560 and its component-only regression coverage. Related kept-service behavior originated in #14544.

## What

Remove the auto-spy prerequisite from runtime dependency selection and retain a snapshot of keep and exclude decisions for declaration and kept-service construction. Keep that snapshot separate from the shared touches provider.

Add regression coverage for classic directives and pipes, field and constructor-body injection, constructor-injection parity, default mock methods, explicit keep/mock/exclude choices, kept modules, and two-argument builder policy. Source tests cover identity, teardown, policy snapshots, and preservation of shared provider state.

## Impact

Default auto-spy mode changes mock method implementations without changing dependency selection. Explicitly kept and excluded root dependencies retain their real implementations in both auto-spy modes, including dynamically created Angular renderers in kept integrations.

Fixes #14899


satanTime:

PR #14907, opened 2026-09-11
fix(MockBuilder): respect parent-only runtime injection #14900

### LINKED ISSUE #14900: Respect injector lookup flags when creating and reusing runtime mocks
## Context

Follow-up to [#14544](https://github.com/help-me-mom/ng-mocks/issues/14544), fixed by [#14557](https://github.com/help-me-mom/ng-mocks/pull/14557). The shared runtime injector was introduced by [#14508](https://github.com/help-me-mom/ng-mocks/pull/14508) and is also used for classic declarations after [#14564](https://github.com/help-me-mom/ng-mocks/pull/14564).

The kept-service fix intercepts root dependency lookups and caches mocks. The interception currently returns a mock without considering the lookup flags, including `SkipSelf`. This can cause a parent-only lookup to return the mock belonging to the current injector.

This is a source-supported defect candidate; the scenario below has not yet been reproduced at runtime.

## Source evidence

- [Cached mocks return immediately](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/libs/ng-mocks/src/lib/common/ng-mocks-runtime-inject.ts#L54), before lookup arguments are considered.
- [Newly created mocks also bypass the original lookup](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/libs/ng-mocks/src/lib/common/ng-mocks-runtime-inject.ts#L69). The original injector receives `...args` only when the lookup is delegated.
- [Existing source units](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/libs/ng-mocks/src/lib/common/ng-mocks-runtime-inject.spec.ts) cover descriptor restoration, caching, and nested active configurations, but not lookup flags.
- [The kept-service regression](https://github.com/help-me-mom/ng-mocks/blob/2e5453697/tests/issue-14544/test.spec.ts) covers root/local injectable targets, constructor parity, and eager initialization, but not parent-only dependency lookup.

## Minimal scenario

With auto-spy enabled, directly keep the following service using `MockBuilder(TargetService)`:

```ts
class RootDependency {
  echo() {
    return 'real';
  }
}

class TargetService {
  local = inject(RootDependency);
  parent = inject(RootDependency, {
    skipSelf: true,
    optional: true,
  });
}
```

Use an environment where the parent injector has no `RootDependency` provider. The ordinary lookup should receive the current injector's mock; the parent-only optional lookup should return `null`. The current cache branch appears to return the same local mock for both.

Reverse the field order to cover a parent-only lookup before any mock has been cached. Also cover an explicit parent provider so the test distinguishes a parent result from the current injector's cached mock.

## Acceptance criteria

- Add a real-Angular regression demonstrating `SkipSelf` behavior before and after a runtime mock is cached, with an absent parent provider and an explicit parent provider.
- Pair it with source-unit coverage showing the appropriate lookup flags and fallback value reach the original injector when delegation is required.
- Preserve same-injector mock identity for ordinary lookups, including subsequent `TestBed.inject()` calls.
- Preserve explicit provider choices and optional missing-provider behavior. Cover supported numeric flags and options-object forms at their actual compatibility boundaries.
- Change only the interception/cache behavior needed to respect injector lookup semantics; retain existing teardown and nested-configuration coverage.

Keep auto-spy policy and framework runtime-manager preservation in their separate follow-ups.

satanTime:
## Why

Runtime injection intercepted root dependency lookups before considering `SkipSelf`. A parent-only lookup could therefore create or reuse the current injector's mock instead of returning an explicit parent provider or Angular's missing-provider result. The same cache shortcut affected later lookups outside service or declaration construction.

Follow-up to #14544 and its kept-service dependency mocking fix.

## What

Respect numeric and options-object `SkipSelf` flags before selecting a cached or new runtime mock. Delegate parent lookups with the original receiver, arguments, and fallback value while preserving the existing suspension of the local mocking scope.

Add source and real-Angular regressions for both lookup orders, absent and explicit parent providers, local mock identity, explicit local providers, optional missing dependencies, custom fallbacks, teardown, and reentrant delegation. Gate the public numeric and options APIs at their supported Angular versions.

## Impact

Parent-only injection follows Angular's injector hierarchy. Ordinary lookups retain the same mock across construction and subsequent injector calls, including `Self`, `Optional`, and `Host` behavior. Existing keep/exclude choices and framework manager protections remain intact.

Fixes #14900


satanTime:

PR #14938, opened 2026-09-11
fix(ngMocks): handle signal form changes and touches #14909

### LINKED ISSUE #14909: Bug: ngMocks.change and ngMocks.touch throw on [formField] (Angular 22 signal forms)
### Description of the bug

Hi, it's been a while! 🙂 Hope all is well.

We're preparing a migration to Angular 22 signal forms, we've only just started, and we're already stuck on the very first interaction: `ngMocks.change()` on a native input bound with `[formField]` fails with:

```
Cannot find ControlValueAccessor on the element. If it is a mock input with [formControlName], you need either to avoid mocking ReactiveFormsModule or to avoid accessing the control in such a way, because this tests ReactiveFormsModule instead of own implementation.
```

`ngMocks.touch()` fails the same way, it goes through the same lookup.

I had a look at the source, and the good news is that the code that would work is already there 🙂

- `FormField` does provide an `NgControl` (its `interopNgControl`), but that `NgControl` has no value accessor: `valueAccessor` is hardcoded to `null` there.
- So `func.get-vca.ts` finds an `NgControl`, finds no CVA, and throws on line 31. It is called from `mock-helper.change.ts:104`.
- Right after, on line 105, there is already the `hasListener(el)` check, and it would be true here: Angular registers `input` and `blur` on that element.
- So `triggerInput()` would do the job, it is just never reached.

Two other cases, in case it helps. A child that keeps a `ControlValueAccessor` fails too, but `NG_VALUE_ACCESSOR` is still on the element injector, so reading it when `NgControl.valueAccessor` is null would bring `__simulateChange` back. A child implementing `FormValueControl` fails as well, but that one has no DOM path at all (its value is a model input), so we're not asking for it here.

### An example of the bug

The sandbox templates aren't on Angular 22 yet, so here is a single spec file instead. Plain `TestBed`, nothing mocked, so it is not a `MockBuilder` side effect. The first test fails, the second passes:

```ts
import { Component, signal } from '@angular/core';
import { FormField, form } from '@angular/forms/signals';

import { MockRender, ngMocks } from 'ng-mocks';

@Component({
  selector: 'target-signal-forms',
  imports: [FormField],
  template: `
    <input [formField]="f.name" />
  `,
})
class TargetComponent {
  public readonly model = signal({ name: '' });
  public readonly f = form(this.model);
}

describe('ngMocks.change with signal forms', () => {
  beforeEach(() =>
      imports: [TargetComponent],
    }),
  );

  it('fails', () => {
    const fixture = MockRender(TargetComponent);

    ngMocks.change('input', 'Ada');

    expect(fixture.point.componentInstance.model()).toEqual({ name: 'Ada' });
  });

  it('passes', () => {
    const fixture = MockRender(TargetComponent);
    const input = ngMocks.find<HTMLInputElement>('input');

    input.nativeElement.value = 'Ada';
    ngMocks.trigger(input, 'input');

    expect(fixture.point.componentInstance.model()).toEqual({ name: 'Ada' });
    expect(fixture.point.componentInstance.f.name().dirty()).toEqual(true);
  });
});
```

Reproduced with ng-mocks 14.17.5 and 14.17.6 on Angular 22.1.5, and on your repo at HEAD.

Expected: `ngMocks.change()` writes the value into the field, like it does on a `formControlName` input. Actual: the error comes before `triggerInput()` is reached, and the model stays unchanged.

Our workaround for now is `input.nativeElement.value = 'Ada'; ngMocks.trigger(input, 'input')`, which writes the model and marks the field dirty (touched needs a `blur` on top). That is also what the Angular team does on their signal forms testing page.

Would you accept a PR for the two first cases? I have a small patch with a regression spec ready, otherwise I'll leave it to you 🙂 Thanks !

satanTime:
## Why

`ngMocks.change()` and `ngMocks.touch()` throw on signal-form fields because `FormField` exposes an `NgControl` whose `valueAccessor` is null. The lookup stops before native listeners can run and misses the `NG_VALUE_ACCESSOR` provider on real or mocked CVA children.

The same fallback lookup can also resolve a parent `FormControlDirective` or `NgModel`, causing an inner signal field to update or touch its ancestor's control.

The signal-forms guide also needs a Docker build that can read linked-worktree Git history for article authors and update dates.

## What

- Allow native listener handling when accessor lookup is optional, while preserving known-control and mock simulation callbacks.
- Recover an element-local value accessor for bound controls, following Angular's selection behavior and preserving an already selected accessor. Keep Angular's optional selector compatible with older Angular consumer bundlers.
- Keep legacy fallback controls local to the element, preserving ancestor values, touched state, and change emissions.
- Add source-unit and Angular signal-form regressions for native text inputs, textareas, selects, real and mocked CVAs, accessor precedence, and nested-field isolation.
- Add a dedicated signal-forms guide with executable native-field and mocked-CVA examples and CodeSandbox/StackBlitz links for both. It covers kept form infrastructure, validation feedback, dirty and touched state, model-to-view updates, and CVA callback wiring.
- Record in the agent runbook and issue-triage skill that documentation changes must teach useful information or correct necessary guidance; a fix alone does not justify obvious capability statements.
- Require live-example links in the docs guidance, accounting for `ng-mocks-sandbox` being updated after release. Keep related articles adjacent in the sidebar, with Signal Forms immediately after Form Controls.
- Add a Docker docs target that reads the current worktree's Git metadata through a read-only mount, preserving article author/date information without mounting primary working files.

## Impact

Existing helper calls work with native signal-form fields and CVA children, including when those fields are nested inside classic form controls. Missing and unsupported accessor diagnostics remain available, and unbound CVAs remain excluded from provider fallback.

Model-only `FormValueControl` support and native checkbox value handling require separate changes to their interaction paths.

Fixes #14909


satanTime:

PR #14939, opened 2026-09-11
fix: mock components from queried containers #14910

### LINKED ISSUE #14910: Check query-first ViewContainerRef dynamic component mocking after #7742
Follow-up to #7742 and [PR #14515](https://github.com/help-me-mom/ng-mocks/pull/14515) (`9706b5274`).

This is a source-supported suspected regression, identified without executable validation. The case needs a failing regression before any implementation change.

## Suspected gap

The defer fix stopped eagerly obtaining the root `ViewContainerRef` on Angular versions exposing `ViewContainerRef.__NG_ELEMENT_ID__`. Dynamic component mocking now installs its concrete `createComponent` patch when that injection hook first executes.

A `ViewContainerRef` obtained from a view or content query follows a different Angular path. `createSpecialToken(..., ViewContainerRef)` calls `createContainerRef` directly; it does not call `ViewContainerRef.__NG_ELEMENT_ID__`. That path is present in the inspected Angular 14 and Angular 22 sources.

Consequently, a query can supply the first `ViewContainerRef` used in the process before ng-mocks patches its concrete prototype. Calling `createComponent(ChildComponent)` can then bypass the configured mock and construct the real child. A previous unrelated test that injects `ViewContainerRef` can hide this problem by installing the patch globally.

Relevant current source:

- [`patchVcrInstance`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts#L381-L405) patches the concrete instance/prototype and uses a process-wide installed flag.
- [`viewContainerInstall`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts#L425-L450) reaches that patch through `__NG_ELEMENT_ID__` on modern Angular.
- Angular's [query special-token resolution](https://github.com/angular/angular/blob/14.3.0/packages/core/src/render3/queries/query.ts) and [ViewContainerRef injection](https://github.com/angular/angular/blob/14.3.0/packages/core/src/linker/view_container_ref.ts) create the same concrete container through separate entry points.

## Minimal scenario

1. A real host contains only `<ng-container #slot></ng-container>` and obtains `slot` using `@ViewChild('slot', {read: ViewContainerRef})`. It does not inject `ViewContainerRef` through its constructor or `inject()`.
2. Its NgModule declares the host and a child whose constructor increments a counter. Configure `MockBuilder(HostComponent, TargetModule)` so the child is mocked.
3. In an isolated first-use run, render the host and call `host.slot.createComponent(ChildComponent)`.
4. Assert the returned component instance is a mock of `ChildComponent`, the real constructor counter remains zero, and the original child's template is absent.

The expected result must be the same whether a separate injected `ViewContainerRef` was used earlier or not.

## Investigation and regression acceptance

- Capture the isolated query-first failure before changing source. A full suite that previously injected a container is insufficient to establish this first-use case.
- Check view and content query acquisition where they share Angular's special-token path; identify the actual affected version boundary.
- Retain the existing constructor-injected dynamic creation behavior, including real children when no mock is configured.
- Preserve [`tests/issue-7742/test.spec.ts`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-7742/test.spec.ts#L50-L65): defer blocks must still be discoverable and render their configured mock. Avoid restoring the eager root lookup that caused that failure.

Existing [`issue-4693`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests-e2e/src/issue-4693/test.spec.ts#L18-L22) obtains its container through constructor injection. [`ng-mocks-render:idea`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/ng-mocks-render/idea.spec.ts) obtains containers through content queries but only calls `createEmbeddedView`. Neither covers query-first dynamic component replacement.

satanTime:
## Why

Angular can create `ViewContainerRef` query results without calling its injection hook. When a view or content query supplied the first container, dynamic component creation bypassed the configured mock and ran the real component's constructor and template.

## What

- Patch containers when Angular refreshes query results, before assigning query setters or returning signal query values.
- Track concrete container prototypes and inherited creation wrappers independently, preserving later instance spies across query refreshes.
- Cover all four decorator and signal query APIs, directive-owned content queries, immediate creation in query setters, independent and inherited container implementations, instance spies, and preservation of query identity and notifications.

## Impact

Dynamic components respect configured mocks whether their container comes from injection or a query. Unconfigured components remain real, and existing injected-container behavior is preserved.

Fixes #14910. Follow-up to #7742 and #14515.


satanTime:

PR #14940, opened 2026-09-11
fix: restore metadata resolution state after failed mocks #14911

### LINKED ISSUE #14911: Check resolver cleanup after failed direct mock construction
Follow-up to #8627 and [PR #14507](https://github.com/help-me-mom/ng-mocks/pull/14507) (`5d4197d56`, merged as `851538adf`).

This is a source-supported suspected cleanup defect, identified without executable validation. The successful circular-import cases remain covered; this follow-up concerns failure during nested mock construction and the next direct mock operation.

## Suspected gap

The circular-import fix correctly removes its new `mockNgDefConstruction` state in a `finally` block. However, nested declaration/module decoration also owns shared resolver and temporary pipe-cache state that is only restored on success:

- [`decorate-declaration.ts`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock/decorate-declaration.ts#L80-L108) creates `mockNgDefResolver` and calls nested `mockNgDef`; its cleanup at [lines 172–174](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock/decorate-declaration.ts#L172-L174) is outside `finally`.
- [`mock-ng-def.ts`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-module/mock-ng-def.ts#L146-L170) pushes resolver state before resolving metadata and only pops/removes it after successful completion.
- [`processMeta`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-module/mock-ng-def.ts#L59-L75) similarly restores the temporary `cachePipe` flag only on success.

If metadata resolution throws after an earlier import has been resolved, a later direct `MockComponent` or `MockModule` call can inherit those stale resolutions. This can matter even after changing a global policy: [`func.global-prepare`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-helper/func.global-prepare.ts#L3-L5) clears completed declaration caches but does not clear this resolver; [`createResolve`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-module/create-resolvers.ts#L70-L87) returns an existing resolver entry before considering the current policy.

`MockBuilder.build()` has its own [outer stash/finally cleanup](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.ts#L79-L117). The direct APIs need their own recovery coverage; a builder-only reproducer could mask this gap.

## Minimal investigation scenario

1. Directly mock a standalone component whose imports resolve `ChildComponent` first and then a module whose provider metadata throws a controlled error while being read.
2. Catch that original error. Assert construction tracking, owned resolver frames, and temporary cache flags have returned to their prior state.
3. Call `ngMocks.globalKeep(ChildComponent)`, then directly mock a different component importing that child.
4. Assert its generated imports contain the real child, and a rendered use of that import uses the real child rather than a stale mock from the failed attempt.

A focused internal regression can force the exception after the first import has populated the resolver, without relying on Angular compilation rejecting malformed metadata. Pair that with the smallest public direct-API recovery case supported by the actual path.

## Expected outcome and acceptance

- Rethrow the original construction error while restoring only the state owned by the failed call.
- Subsequent direct mock creation must respect the current keep/mock policy; the failed call must not permanently enable pipe caching.
- Preserve a pre-existing outer resolver during a nested failure and retain completed mock caching where intended.
- Keep the original direct self-import and two-component cycle cases in [`tests/issue-8627/test.spec.ts`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-8627/test.spec.ts#L35-L52), plus the compiled circular import rendering in [`tests/issue-6143/test.spec.ts`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-6143/test.spec.ts).

The new construction Set already has correct local cleanup in [`get-mock.ts`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock/get-mock.ts#L25-L49). Investigation should focus on the remaining shared state and avoid replacing that working cycle guard.

satanTime:
## Why

Failed direct mock construction could leave shared metadata resolutions or temporary pipe caching behind. A later mock could then ignore an updated keep policy or reuse an earlier pipe transform. Reviewing the cleanup introduced for #8627 also found that `ngMocks.guts` could leak its local policies after a failure or discard an outer caller's state.

## What

- Restore temporary resolver and pipe-cache state when metadata or provider resolution throws, including nested modules, standalone imports, and host directives.
- Restore the outer resolver and dependency policies around `ngMocks.guts`, while preserving completed mock caches and propagated resolutions.
- Add regression and preservation coverage for failed construction, same-test recovery, provider and view-provider errors, independent pipe transforms, module wrappers with providers, and caller-owned state.

## Impact

Later mock construction uses the current dependency policy and each explicit pipe transform remains independent. Successfully completed mocks remain reusable. The change restores the documented API behavior, so public documentation changes are unnecessary.

Closes #14911.


satanTime:

PR #14941, opened 2026-09-11
feat(ngMocks): support signal model controls and correct native form updates

satanTime:
## Why

The signal-form support in #14938 exposed adjacent gaps in `ngMocks.change` and `ngMocks.touch`. Native checkbox, numeric, date, and select controls read browser state that a temporary `value` property did not update. Unbound descendants could also resolve an ancestor's selected value accessor. Custom signal controls communicate through model and touch outputs and have no CVA callback for these helpers to invoke.

The wider regression coverage also exposed a test-isolation defect: Angular 14 can leave a recompiled directive factory behind after a mocked provider override is reset. A later real select then loses its inherited constructor dependencies.

## What

- Update native control state before dispatching events, preserving raw event values where existing callers rely on them and keeping subsequent browser and Angular writes usable.
- Require a local `NgControl` provider when resolving its selected accessor, while retaining selected-CVA and legacy form fallback precedence.
- Support real and mocked `FormValueControl` and `FormCheckboxControl` dependencies through their declared outputs, including Angular 21's touched model and Angular 22's touch output.
- Restore original declaration factories after provider overrides are reset, including cleanup failures, while preserving the original teardown error and constructor injection for later tests.
- Cover native controls, ancestor isolation, null and missing local providers, model identity, parent-to-child feedback, untouched sibling fields, and event side effects in source-unit and compiled Angular regressions. Declarations without bindings cannot displace a co-located model control.
- Preserve older DOM runners that lack numeric/date accessors and align the Vitest DOM Date constructor so Angular recognizes dates returned by native inputs.
- Extend the signal-forms guide with executable specs for a mocked model control and both live-example links. The sandbox receives those examples after release.

## Impact

Form tests can interact with native, CVA, and signal-model controls through the same helpers. Changing a custom signal control updates its value and dirty state; touching it remains a separate interaction. Existing CVA behavior retains priority.

Follow-up to #14938. Related to #14909.


satanTime:

PR #14942, opened 2026-09-11
fix(TestBed): complete teardown cleanup after errors #14912

### LINKED ISSUE #14912: Check TestBed recovery after provider teardown throws
## Context

Follow-up to #9666, #5437 and #11324.

Existing fixes ensure that flush/faster paths invoke TestBed teardown and respect teardown configuration. The remaining risk is recovery when a kept provider's `ngOnDestroy` throws.

## Suspected wider defect

[`mock-helper.flush-test-bed.ts`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/libs/ng-mocks/src/lib/mock-helper/mock-helper.flush-test-bed.ts#L9) calls `tearDownTestingModule()` before clearing `_instantiated`, `_moduleFactory` and `_testModuleRef`, without a `finally` block. If Angular rethrows a teardown error, these resets are skipped and the destroyed module reference can remain attached to TestBed.

Angular's own `resetTestingModule()` clears its module reference in `finally` even when teardown throws. The existing flush helper tests replace teardown with spies; #5437 covers normal provider destruction and #11324 covers disabled teardown. They do not establish recovery after a destruction error.

This is a source-supported failure-path finding; a runtime reproduction is still required.

- Keep a provider whose `ngOnDestroy` throws once. Capture the flush error, then create another fixture and inject the provider again; require a fresh instance without a destroyed-injector error or a repeated destruction attempt on the old module.
- Exercise explicit flush and the relevant faster/reset caller, keeping the regression isolated from other tests.
- Preserve Angular's `rethrowErrors` policy, normal one-time destruction, and `destroyAfterEach: false` behavior.
- Add focused source-unit assertions that cleanup fields and injection seed state are reset on exceptional teardown, plus real-Angular lifecycle coverage.
- Fix shared cleanup ordering if the reproduction confirms the stale state; do not swallow an error that Angular is configured to rethrow.

satanTime:
## Why

A provider teardown error could leave a destroyed injector attached to TestBed, preventing later rendering. Pending ng-mocks overrides could also stop Angular's own reset, leaving compiler metadata behind. Reviewing the related faster cleanup found that fixture errors interrupted remaining cleanup and finishing an inner faster scope could destroy an outer scope's fixture.

## What

- Clear TestBed initialization state after teardown, including failures, while retaining Angular's teardown configuration.
- Complete pending override restoration and Angular reset while preserving the first error.
- Finish faster fixture cleanup after errors and keep outer faster scopes alive until their own cleanup.
- Add regressions for flush and render retries, overridden declarations, nested fixtures, error precedence, provider destruction, and disabled or suppressed teardown.

## Impact

Tests can recover after a caught teardown error without reusing a destroyed injector or stale compiler metadata. Existing error policies and retained fixtures remain covered. This restores expected behavior and requires no public documentation changes.

Closes #14912.

### DIFF
--- a/libs/ng-mocks/src/lib/common/func.apply-test-bed-override.ts
+++ b/libs/ng-mocks/src/lib/common/func.apply-test-bed-override.ts
@@ -1,0 +1,18 @@
+import { TestBed } from '@angular/core/testing';
+
+import { isNgDef } from './func.is-ng-def';
+
+export default (def: any, override: any): void => {
+  if (isNgDef(def, 'c')) {
+    TestBed.overrideComponent(def, override);
+  } else if (isNgDef(def, 'd')) {
+    TestBed.overrideDirective(def, override);
+  } else if (isNgDef(def, 'm')) {
+    TestBed.overrideModule(def, override);
+  }
+  if (isNgDef(def, 't')) {
+    TestBed.overrideProvider(def, override);
+  } else if (isNgDef(def, 'i')) {
+    TestBed.overrideProvider(def, override);
+  }
+};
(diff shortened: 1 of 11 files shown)


satanTime:

PR #14943, opened 2026-09-11
feat(MockRender): preserve selected callable params #14913

### LINKED ISSUE #14913: Check callable input data in custom MockRender templates after #13089
## Background

Follow-up to [#13089](https://github.com/help-me-mom/ng-mocks/issues/13089) and [#14491](https://github.com/help-me-mom/ng-mocks/pull/14491), which preserve callable component input values, including Angular Signal Forms `FieldTree` proxies.

The fix derives public input names from the component passed to `MockRender` or `MockRenderFactory`. The adjacent custom-template route has no component metadata. It still passes an empty value-key list to the property reader, even when an explicit factory binding is used as a component input. The reader consequently clones a callable data parameter as though it were a callback. The clone copies named properties but does not preserve arbitrary proxy behavior or symbol properties.

Relevant current paths: [factory installation](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-render/mock-render-factory.ts#L258), [callable property reader](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-render/func.install-prop-reader.ts#L11), and [function clone](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/helper.create-clone.ts#L6).

This is a source-supported risk in an adjacent public route; the custom-template runtime failure still needs a focused reproducer.

## Bounded investigation

Use the same real `FieldTree` input as the [existing regression](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-13089/test.spec.ts#L51), but render `<target-13089 [items]="items"></target-13089>` with both `MockRender(template, params)` and `MockRenderFactory(template, ['items'])`.

- Confirm that the child receives usable callable data, can iterate the field tree, and renders updated form state. Check whether the parameter's identity and symbol/proxy capabilities survive.
- Add a classic function-valued input case at its older compatibility boundary, without requiring Signal Forms.
- Preserve callback receiver binding, clone caching, and callback replacement. These are intentional behaviors already covered by [the custom-template callback regression](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/mock-render-factory/zoneless.spec.ts#L83).
- Preserve explicit binding selection and zoneless scheduling established by #14898.

Change production behavior only after demonstrating the missing route. Do not simply stop rebinding every custom-template function: callbacks and callable input data have different requirements. If an explicit way to identify callable data is needed, document that contract rather than inferring support from the binding list alone.

This concerns callable data transport, not conditional/default factory bindings requested by #10423.

It is also separate from #14909: that report concerns `ngMocks.change`/`touch` locating a value accessor for `[formField]`. This case concerns transporting a callable value through the render wrapper before any form interaction helper is used.

satanTime:
## Why

Custom MockRender templates cannot distinguish a callable input value from a callback. The callback wrapper can change function identity and lose proxy behavior, making values such as Signal Forms FieldTree unusable. The automatic input handling from #13089 covers declaration-based rendering, but a custom template needs an explicit way to identify data values.

## What

- Add `valueKeys` to MockRender and MockRenderFactory options to preserve selected params values by identity, alongside automatically recognized declaration inputs.
- Carry that policy through service and provided-pipe rendering, which install a second property reader.
- Cover callable component and directive inputs, FieldTree rendering and updates, shared factory caches, replacement writes, aliases, callback receivers, and zoneless binding selection.
- Document the option with an executable example combining callable input data and an output callback.

## Impact

Custom templates can pass callable data without cloning it. Existing calls retain their callback behavior, declaration inputs remain automatic, and `valueKeys` does not change binding selection or change-detection scheduling.

Fixes #14913.
Follow-up to #13089 and #14491.


satanTime:

PR #13921, opened 2026-05-24
fix(a22): update angular monorepo

satanTime:
Split from #13347 so the Renovate PR can stay dependency-only while this PR carries the compatibility fix and can run the broader e2e matrix.

What changed:
- Keeps the Angular 22 ComponentFactoryResolver value import optional so ng-mocks still loads when the export is removed.
- Keeps MockRender wrappers checkable under Angular 22 while preserving older Angular behavior.
- Restores CVA and validator proxy handling across Angular forms versions.
- Adds comments and focused coverage for the compatibility paths.
- Leaves the upstream zone.js version unchanged.

Validation run locally:
- sh test.sh coverage
- sh test.sh a5
- sh test.sh a14
- sh test.sh a20
- sh test.sh a22
- npm run ts:check
- npm run prettier:check
- git diff --check

Issue #13854, opened 2026-05-14
Bug: NG0202 'dependency at index 1 of the parameter list is invalid' under Angular 22 next.10 + TypeScript 6


ghaiat-yoobic:
### Description of the bug

Auto-mocked services with `inject()` field initialization throw `NG0202: This constructor is not compatible with Angular Dependency Injection because its dependency at index 1 of the parameter list is invalid` during `TestBed.inject()` / `TestBed.finalize()`. The thrown frame shows ng-mocks's generated `MockMiddleware<N>_Factory` calling `ɵɵinvalidFactoryDep(1)`.

Reproduces on every spec that uses `MockBuilder(Service).mock(...)` against an `@Injectable({providedIn:'root'})` class whose dependencies are declared via `inject()` field initializers (the modern Angular 17+ idiom).

### Versions

- `ng-mocks`: 14.15.2
- `@angular/core`: 22.0.0-next.10 (peer requires `typescript >=6.0 <6.1`)
- `typescript`: 6.0.3
- `jest`: 30.3.0
- `jest-preset-angular`: 16.1.4
- `@types/jest`: 30.0.0
- DOM env: `@happy-dom/jest-environment` 20.x (also reproduces with jsdom)

For context, ng-mocks's own `e2e/a22` validated combo pins `typescript: 5.9.3` and `@angular/core: 22.0.0-next.3`. PR #13383 (open since 2026-03-23) bumps TS to 6.0 and is auto-merge-enabled but hasn't landed.

### Repro shape

```ts
// some.service.ts
@Injectable({ providedIn: 'root' })
export class SomeService {
  private a = inject(SomeDep);
  private b = inject(AnotherDep);
  private c = inject(ThirdDep);
  // ... methods
}

// some.service.spec.ts
describe('SomeService', () => {
  ngMocks.faster();
  beforeAll(() => MockBuilder(SomeService));

  it('should create', () => {
    const service = TestBed.inject(SomeService);
    expect(service).toBeTruthy();
  });
});
```

### Stack

```
NG0202: dependency at index 1 of the parameter list is invalid
  at ɵɵinvalidFactoryDep (@angular/core/.../_pending_tasks-chunk.mjs:687:9)
  at Object.MockMiddleware<N>_Factory [as useFactory] (ng:/MockMiddleware<N>/ɵfac.js:6:7)
  at Object.factory (@angular/core/.../_pending_tasks-chunk.mjs:1395:32)
[...]

PR #14944, opened 2026-09-11
fix: isolate inherited nonconfigurable Angular definitions #14914

### LINKED ISSUE #14914: Cover inherited Ivy definition isolation across generated declarations
## Context

Follow-up to #13854 and #13921.

The Angular 22 constructor-injection failure was fixed by shadowing inherited Ivy definitions before decorating a generated subclass. The shared `extendClass` helper is also used for selectorless directive/component middleware and generated modules.

## Coverage to add

The [shared helper](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/libs/ng-mocks/src/lib/common/core.helpers.ts#L114) handles component, directive, factory, injector, module and pipe definitions. Its focused unit suite currently tests construction fallbacks and mock names, but does not assert isolation of these inherited definitions.

- Add focused source-unit coverage for inherited definitions, absent definitions, and preservation of the original class's metadata. Verify separately generated children do not share newly decorated definitions.
- Add real-Angular regressions for selectorless component and directive middleware with inherited constructor dependencies. For the module path, use a kept module whose changed dependency forces a cloned subclass, or the decorated entry-components module path, and verify that the generated class actually extends the decorated original. Ordinary fully mocked modules extend the generic mock base and would not establish inherited-definition isolation.
- Assert the rendered declaration and injected provider identities, correct constructor arguments, and that rendering/mocking a child leaves the original declaration usable afterward.
- Exercise the affected Angular compiler/runner boundary, including Angular 22, while preserving supported older declaration paths.

This is a regression-coverage gap identified by source review, not a reproduced new runtime failure. Change source only if these cases expose another instance of the defect.

satanTime:
## Why

The inherited-definition fix in #13921 uses `coreDefineProperty` to shadow Ivy metadata before decorating a generated subclass. The shared descriptor helper also rejects inherited nonconfigurable properties, even though the child can legally define an independent own property. This leaves the original definition visible on the child and also prevents `ngMocks.stub` from shadowing inherited locked values or accessors.

Follow-up to #13854. Fixes #14914.

## What

Check the target's own descriptor before defining a property. Keep both helper signatures and the existing configurable/writable defaults, and continue protecting locked own properties, including those on objects with a null prototype.

Cover all six Ivy definition fields, parent getters, absent fields, sibling and grandchild isolation, selectorless component/directive injection, and kept modules whose replaced declarations require subclasses. Add public stub regressions for inherited locked values and accessors, with assertions that ancestors and siblings remain unchanged and getters are not invoked during replacement.

Prefer existing library helpers for compatible fixtures. Update the runbook and triage skill to investigate related defects while requiring a demonstrated need for additional entry points, parameters, options, abstractions, or public documentation.

Require review of repository security-tool findings and verified GitHub Advanced Security comments during CI, and resolution of valid findings on the current head. Correct the flagged missing-target test to assert that the helper does not throw, without consuming its void return value.

## Impact

Generated subclasses and descriptor-copying callers can shadow native inherited locks without changing the ancestor. Existing entry points and helper defaults remain intact. Public documentation changes are unnecessary because the fix restores the existing contract. Legacy Zone.js's separate inherited lock-bookkeeping limitation is outside this change.

### DIFF
--- a/.agents/skills/triage-issue/SKILL.md
+++ b/.agents/skills/triage-issue/SKILL.md
@@ -89,6 +89,8 @@ Create and maintain a plain Markdown checklist:
    - Correct the confirmed shared cause and its affected paths. Preserve adjacent cases that already work and
      avoid speculative source changes for cases that only need regression coverage.
    - Prefer existing ng-mocks helpers and patterns over new abstractions.
+   - Avoid expanding functionality while fixing wider defects; use existing entry points and defaults unless
+     a demonstrated requirement needs an addition, following `Fix Scope Review` in `AGENTS.md`.
    - Add code comments only for non-obvious Angular behavior, compatibility constraints, or private API handling.
    - Do not hide failures with skips, broad version exclusions, relaxed assertions, or coverage ignores unless the issue truly cannot be represented otherwise.
    - Use functional tests during investigation and implementation. Do not run formatting/Prettier, ESLint, or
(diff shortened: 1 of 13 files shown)


satanTime:

PR #14946, opened 2026-09-11
fix(MockRender): preserve middleware change detection #14915

### LINKED ISSUE #14915: Check implicit OnPush preservation in MockRender middleware clones
## Background

Follow-up to [#14157](https://github.com/help-me-mom/ng-mocks/issues/14157) and [#14489](https://github.com/help-me-mom/ng-mocks/pull/14489), which preserve Angular 22's implicit OnPush behavior when `changeDetection` is omitted.

The fix reads the effective strategy from the compiled component definition because Angular 22's decorator annotations do not distinguish implicit OnPush from explicit Eager. The existing regression renders components with simple element selectors.

Selectorless components and components with complex or multiple selectors take a different route: [MockRender recompiles a middleware component from reflected metadata](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-render/func.reflect-template.ts#L44), supplying Default before spreading that metadata. The [compiled-definition inheritance helper](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-render/func.inherit-definition.ts#L3) restores host/animation metadata but does not restore `onPush`.

This suggests that a real component with an omitted strategy can become eager when a middleware clone is required. It is a source-supported suspicion; the compiled clone's runtime behavior needs a direct regression.

On Angular 22, repeat the [existing implicit-strategy assertions](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-14157/test.spec.ts#L33) for a selectorless component and a component whose complex selector requires cloning:

- Render an array input, mutate the array without replacing its reference, and call `fixture.detectChanges()`. The view should retain the previous value under implicit OnPush.
- Replace the input reference and confirm the view updates.
- Keep explicit Default/Eager and explicit OnPush controls, including an effective TestBed strategy override if the fix changes override handling.
- Check that the original component's metadata/definition is preserved rather than modified to make its clone pass.

The [existing middleware metadata unit](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-render/func.reflect-template.spec.ts#L217) covers explicit OnPush, which survives annotation copying. It does not establish the omitted-strategy case.

Change source only if the compiled regression fails. Preserve older Angular defaults and the wrapper's own checking behavior. The required behavior is preserving the rendered component's effective strategy across both simple-selector and middleware-clone routes.

satanTime:
## Why

Selectorless components and components with complex or multiple selectors need a middleware clone in `MockRender`. Angular 22's reflected annotations do not distinguish implicit OnPush from explicit eager checking, so recompiling those annotations gives the clone the wrong strategy. The same stale strategy can persist when TestBed overrides the original after a factory creates its clone.

Follow-up to #14157 and #14489. Fixes #14915.

## What

Preserve the original component's effective compiled change-detection strategy through the existing definition-inheritance path. Its existing refresh handles original and clone definitions replaced by TestBed, including overrides that switch either strategy on or off.

Add regressions for selectorless, complex, and multiple selectors; explicit strategies; early and late overrides; and a classic NgModule component across supported Angular versions. Separate Angular 22's implicit default from both late override directions on Ivy starting with Angular 9. Cover nonconfigurable getters, absent strategy metadata, original definitions and annotations, input identity, and subsequent original-component use.

## Impact

OnPush clones retain their view after a same-reference input mutation and update when the input reference changes. Explicit eager checking, older Angular defaults, and the wrapper's checking behavior are preserved. The fix restores the documented behavior, so public documentation changes are unnecessary.


satanTime:

PR #14948, opened 2026-09-12
fix(MockService): mock own accessors without invoking them #14916

### LINKED ISSUE #14916: Check own accessor side effects when mocking explicit object shapes
## Background

Follow-up to [#10919](https://github.com/help-me-mom/ng-mocks/issues/10919), its [own-property change](https://github.com/help-me-mom/ng-mocks/commit/220c71f20eba01e1017d4421a8cf8f6863de8a77), and the subsequent [constructor-suppression correction](https://github.com/help-me-mom/ng-mocks/pull/14554) for #14528.

The current supported approach is to provide an explicit object shape or an intentionally constructed instance to `MockService`. Class mocking must not execute the original constructor merely to discover fields.

The adjacent object path still [enumerates keys and reads `service[property]`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/mock-service.ts#L33). An enumerable own accessor is therefore evaluated while creating the mock. A lazy API getter can perform real work or throw before it is replaced by a mock. Prototype accessors take a separate descriptor-based path and already have coverage.

This is a source-supported risk; an own-accessor regression should establish the exact behavior before a production change.

## Bounded investigation and preservation

- Pass an explicit object with an enumerable own getter that increments a counter or throws to `MockService`. Also pass `MockService(shape)` as a nested value in explicit `MockProvider` overrides, following [the current #10919 example](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-10919/test.spec.ts#L40). This does not assume that MockProvider recursively mocks override values by itself.
- Verify mock construction does not execute the real accessor. Mocked reads/writes should use the established mock-accessor behavior, and an explicit getter stub should remain possible.
- Preserve ordinary own data properties, recursive nested method mocking, and auto-spy behavior, covered by [the object-shape source unit](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/mock-service.spec.ts#L265).
- Preserve prototype getter/setter behavior, covered by [the existing accessor unit](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/mock-service.spec.ts#L310), and preserve the no-constructor guarantee from #14528.

Do not execute a getter to discover the type of its return value. Configure the desired returned shape explicitly when it cannot be recovered from descriptors. This does not request automatic discovery of ordinary signal/Observable fields, which is tracked by #2701, or automatic class initialization.

satanTime:
## Why

Mocking an explicit object shape reads its enumerable own properties to discover nested values. For accessors, that executes the real getter during mock construction and can trigger side effects or throw. Setter-only properties are omitted instead of receiving the usual mock accessor behavior.

## What

Inspect own property descriptors and reuse the existing getter/setter mocking helpers. Add regressions for getter-only, setter-only, paired and shadowing accessors, explicit nested `MockService` overrides in `MockProvider`, and mocked `useValue` providers.

Record agent guidance to prefer precise types over `any`, and use an explicit optional global shape in the new runner setup.

## Impact

Accessor mocks start undefined, retain assigned mock values and accept explicit getter stubs without calling the original accessors. Nested method auto-spies, original descriptors and constructor suppression remain intact. This restores the existing mock behavior without new API parameters or public documentation changes.

Fixes #14916.

### DIFF
--- a/.agents/skills/triage-issue/SKILL.md
+++ b/.agents/skills/triage-issue/SKILL.md
@@ -89,6 +89,7 @@ Create and maintain a plain Markdown checklist:
    - Correct the confirmed shared cause and its affected paths. Preserve adjacent cases that already work and
      avoid speculative source changes for cases that only need regression coverage.
    - Prefer existing ng-mocks helpers and patterns over new abstractions.
+   - Follow `AGENTS.md`'s typing guidance: avoid `any` where practical, including in regression tests.
    - Avoid expanding functionality while fixing wider defects; use existing entry points and defaults unless
      a demonstrated requirement needs an addition, following `Fix Scope Review` in `AGENTS.md`.
    - Add code comments only for non-obvious Angular behavior, compatibility constraints, or private API handling.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -311,6 +311,10 @@
 
 - Prefer existing library helpers when their semantics fit the task, including in tests; for example, use
   `coreDefineProperty` for compatible property definitions.
+- Make a best effort to avoid `any`, including in tests. Prefer inference, concrete types, or narrower object
+  shapes; consider `unknown` with narrowing or `never` when their semantics fit and the target TypeScript version
+  supports them. Try a different declaration or approach before accepting `any`, and do not use casts merely to
+  hide a type error.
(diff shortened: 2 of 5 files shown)


satanTime:

Issue #14949, opened 2026-09-12
Check kept root provider recipes when JIT annotations omit metadata


satanTime:
## Problem

The scope review for #14917 found a separate provider-retention gap when a class has only runtime JIT `Injectable` annotations. Those annotations do not retain the decorator's provider options, so `collectDeclarations` cannot recover the recipe. This affects the existing `useFactory` path as well as `useClass`, `useExisting`, and `useValue`.

Explicitly keeping such a token can register the token class and execute its constructor instead of Angular's configured root provider recipe. Published libraries with retained `decorators` metadata follow the reflection path addressed by #14917; the missing-metadata case needs a separate solution.

For example, a JIT-only fixture with `@Injectable({ providedIn: 'root', useFactory: () => ({ name: 'factory' }) })` and a token class whose constructor throws should resolve the factory value after `MockBuilder().keep(Target)`. The fallback class provider instead runs the constructor.

## Investigation and coverage

- Reproduce the missing-metadata case with real Angular injection, including the existing factory form, aliases, implementations, and object or empty values.
- Preserve module-specific provider recipes and explicit `.provide(...)` precedence. Replacing every ordinary kept class with a native-factory object changes the existing `builtProviders.get(token) === token` precedence path.
- Check inherited injectable definitions: copying an inherited factory directly must not instantiate the parent when Angular would construct the child.
- Preserve automatic dependency mocking and explicit keeps or exclusions when a native factory resolves hidden `inject()` dependencies.
- Cover retained-library metadata as a preservation case and use the actual Angular API boundaries. Avoid adding public options or new entry points for an internal provider-retention fix.

Related to #7938 and #14917.

satanTime:
[v14.18.0](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.18.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

Issue #7938, opened 2024-01-13
Bug: `.keep(ReactiveFormsModule)` does not actually keep


JPtenBerge:
Hello there!

### Description of the bug

During testing, I'm trying to keep the `ReactiveFormsModule` from being mocked using `.keep()`, but it very much seems to be mocked and then causing errors because it's a seemingly shallow mock.

### An example of the bug

I have this component where I'm initializing a reactive form in the `ngOnInit()`:

```ts
@Component({
  selector: 'app-root',
  standalone: true,
  imports: [ReactiveFormsModule],
  templateUrl: './app.component.html',
})
export class AppComponent implements OnInit {
  form!: FormGroup<AddPersonForm>;

  constructor(private fb: NonNullableFormBuilder) {}

  ngOnInit() {
    this.form = this.fb.group<AddPersonForm>({
      name: this.fb.control(''),
      age: this.fb.control(21),
    });
  }
}
```

I'm trying to test it, where I actually want to keep the `ReactiveFormsModule` untouched, it should just work as in production:

```ts
describe('AppComponent', () => {
  let sut: AppComponent;

  beforeEach(() => MockBuilder(AppComponent).keep(ReactiveFormsModule));

  it('should initialize the form without errors', () => {
    sut = MockRender(AppComponent).componentInstance;
    sut.ngOnInit();
  });
});
```

But when running the test, it doesn't seem to keep that module as I intend:

```sh
TypeError: this.fb.control is not a function
```

If I `console.log(this.fb);` in my `AppComponent`, I see a mock:

```sh
NonNullableFormBuilder
   __ngMocks: true
   [[Prototype]]: Object
```

Link: https://github.com/JPtenBerge/ng-mocks-reactive-forms-bug

### Expected vs actual behavior

Expected: all form functions (`fb.group()`, `fb.control()`) to work as when serving the app.
Actual: it's a mock without any of the functions.

felixdenis:
I have encountered the same problem. 
@JPtenBerge Try explicitly keeping NonNullableFormBuilder in addition to ReactiveFormsModule.

felixdenis:
I believe FormBuilder and its variants are not provided by FormsModule or ReactiveFormsModule. Hence this is not a bug in ng-mocks.

JPtenBerge:
> I have encountered the same problem. @JPtenBerge Try explicitly keeping NonNullableFormBuilder in addition to ReactiveFormsModule.

Have tried this in various ways:

[...]

PR #14950, opened 2026-09-12
fix(MockBuilder): preserve kept root provider recipes #14917

### LINKED ISSUE #14917: Check non-factory provider recipes on explicitly kept root injectables
## Context

Follow-up to [#7938](https://github.com/help-me-mom/ng-mocks/issues/7938) and its [root factory preservation fix](https://github.com/help-me-mom/ng-mocks/commit/86cb0e962ee9dcd9e67633cc8c13f437de82997a).

The original fix preserves `useFactory` when a root injectable is explicitly kept. Other provider recipes declared in `@Injectable` metadata enter the same path but are not recognized by that helper.

## Suspected wider defect

[`getRootProviderKeepProvider`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/libs/ng-mocks/src/lib/mock-builder/promise/get-root-provider-keep-provider.ts#L8) recognizes `Injectable.useFactory` and `Service.factory`. For `useExisting`, `useClass`, or `useValue`, it returns `undefined`. [`initKeepDef`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/libs/ng-mocks/src/lib/mock-builder/promise/init-keep-def.ts#L14) then stores the class itself, and the generated testing module registers that class provider.

An explicit class provider uses the token's class factory, so this can lose the root injectable's configured alias, implementation, or value. For example, explicitly keeping an injectable declared with `@Injectable({providedIn: 'root', useExisting: Implementation})` should retain the alias rather than instantiate the alias token's own class.

This is a source-supported likely defect; a runtime reproduction is still required. It concerns explicit `.keep()` semantics, separate from the default root-provider policy discussed in #4701.

- Add real Angular regressions for explicitly kept root injectables whose metadata uses `useExisting`, `useClass`, and `useValue`.
- For `useExisting`, also keep the implementation and assert that both tokens resolve to the same instance. For `useClass`, assert the configured implementation and dependency identity. For `useValue`, assert exact configured identity and a supported falsy-value case.
- Assert that the token classes' own constructors do not run when the metadata selects another recipe.
- Preserve the existing `useFactory`/`deps` behavior and explicit `.provide()` precedence. Add focused helper assertions that reflecting these recipes does not mutate the original metadata.
- Correct the shared helper only after reproducing the missing behavior, with version gates matching the APIs used by the fixtures.

The existing [#7938 regression](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/tests/issue-7938/test.spec.ts#L37) and helper units cover factory providers. The `TestProviderWithUseExisting`, `TestProviderWithUseClass`, and `TestProviderWithUseValue` examples configure ordinary module providers; they do not cover these recipes in root injectable metadata.

satanTime:
## Why

Explicitly keeping a root injectable reconstructed its factory recipe, but dropped reflected `useExisting`, `useClass`, and `useValue` recipes. Angular could then instantiate the token class instead of resolving its alias, implementation, or configured value. This extends the provider-metadata handling introduced for #7938.

JIT-only annotations that omit recipe metadata are tracked separately in #14949.

## What

Preserve the remaining reflected provider recipes, including value identity, empty values, forward references, and explicit or inferred class dependencies. Retained class recipes use their existing Angular factory inside the runtime mocking window when `MockBuilder` should mock hidden `inject()` dependencies.

Add focused helper coverage and real Angular regressions using the retained decorator metadata emitted by compiled libraries. Cover both explicit keeps and builder targets, existing factory behavior, dependency identities, and explicit-provider precedence.

## Impact

Kept library providers resolve according to their metadata while existing automatic dependency mocking and explicit overrides remain intact. This restores the expected contract without adding public options or requiring documentation changes.

Fixes #14917


satanTime:

Issue #9655, opened 2024-08-03
Bug: Structural directive inheritance breaks


nsgundy:
### Description of the bug

Unit tests for a structural directive break when `ng-mocks` is used anywhere in a test (even if that test is unrelated to that directive).

This is the structural directive:
```ts
@Directive({
  selector: '[appSomeHeader]',
  standalone: true,
  inputs: [
    {name: 'foo', alias: 'appSomeHeader'},
  ],
})
export class SomeHeaderDirective extends BaseHeaderDirective {
  @Input({alias: 'appSomeHeaderBar', transform: booleanAttribute}) override bar: boolean = false;
}
```

The `BaseHeaderDirective` being extended:
```ts
@Directive({
  selector: '[appBaseHeader]',
  inputs: [{name: 'foo', alias: 'appBaseHeader'}],
  standalone: true
})
export class BaseHeaderDirective {
  foo: string = '';
  @Input({alias: 'appBaseHeaderBar', transform: booleanAttribute})
  bar: boolean = false;
}
```

The test being run:
```ts
describe('SomeHeaderDirective', () => {
  beforeEach(waitForAsync(() => {
    TestBed.configureTestingModule({
      imports: [HostComponent],
    }).compileComponents();
  }));

  it('should have correct foo and bar value', () => {
    const fixture = TestBed.createComponent(HostComponent);
    fixture.detectChanges();
    const debugElement = fixture.debugElement.queryAllNodes(By.directive(SomeHeaderDirective))[0];
    const directive = debugElement.injector.get(SomeHeaderDirective);
    expect(directive.foo).toBe('foo');
    expect(directive.bar).toBe(true);
  });
});
```

The above test works as expected, unless we introduce `ng-mocks` into the project. It breaks, even if `ng-mocks` not actively being called. Example code that will break the above test (can by in any `.spec.ts` file):
```ts
  afterEach(() => {
    if (false) {
      ngMocks.flushTestBed();
    }
  })
```

The error shown:
```
Error: NG0303: Can't bind to 'appSomeHeaderBar' since it isn't a known property of 'div' (used in the 'HostComponent' component template).
```

### An example of the bug

Created a minimal reproduction here:
https://github.com/nsgundy/ng-mocks-structural-directive-inheritance-issue

### Context

[...]

PR #14951, opened 2026-09-12
test(TestBed): cover inherited structural input aliases #14918

### LINKED ISSUE #14918: Retain inherited structural-input alias regression coverage
## Context

Follow-up to [#9655](https://github.com/help-me-mom/ng-mocks/issues/9655).

The report described a structural directive with inherited input aliases and a transform, where loading ng-mocks affected an otherwise ordinary TestBed test. The final recheck passed on current main, including the variant where a derived directive's `inputs` array overrides a base property's decorated alias. There is no retained regression for that exact setup.

## Coverage to add

Current [metadata collection](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/libs/ng-mocks/src/lib/resolve/collect-declarations.ts#L55) clones inherited input lists and combines class-array, decorator, and compiled metadata. The newer [input-transform suite](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/tests/mock-render-metadata/input-transforms.spec.ts) covers related precedence cases through MockRender/MockComponent. It does not lock down inherited property-to-class-array alias overrides in structural microsyntax under ordinary TestBed.

- Retain the reported base directive with a transformed `@Input`, a derived class-array alias override, and the host's `*appSomeHeader="'foo'; bar: true"` binding.
- Configure and render the host using ordinary TestBed with ng-mocks loaded. Assert that the derived aliases bind and that the resulting input values and transform result are correct.
- Then render a host using the base declaration's original aliases, and repeat the relevant fixture/setup cycle. Assert that neither derived reflection nor TestBed interception mutates the base input metadata or aliases.
- Gate the regression at the actual input-transform API boundary, and establish it in an Angular-compiled target.

This is regression protection for a previously resolved report, not a claim that current runtime behavior is broken. Investigate and fix any shared metadata defect only if the retained case exposes one.

satanTime:
## Why

The inherited structural-directive setup from #9655 now works, but lacked a retained regression. Its derived `inputs` array overrides an alias declared by a transformed input on the base directive, and the failure originally affected ordinary TestBed tests merely because ng-mocks was loaded.

## What

Retain the reported microsyntax binding in ordinary TestBed and repeat it after explicitly reflecting the derived directive. Exercise true and string-false inputs, then render the base using its original aliases across repeated setup/reset cycles. Assert that the original input arrays and decorated input options remain unchanged.

Gate the regression at the input-transform API boundary, Angular 16.1 (the repository's Angular 16 target uses 16.2).

## Impact

Protect the existing behavior and metadata isolation without changing library functionality. Public documentation does not need an update.

Fixes #14918

### DIFF
--- a/test-spread.conf
+++ b/test-spread.conf
@@ -136,6 +136,7 @@ file|examples/TestContentChild/signals.spec.ts|features=signalQueries
 file|tests/mock-render-metadata/host-directives.spec.ts|features=hostDirectives
 file|tests/mock-render-metadata/static-queries.spec.ts|versions=8+
 file|tests/mock-render-metadata/input-transforms.spec.ts|versions=16+
+file|tests/issue-14918/test.spec.ts|versions=16+
 file|tests/mock-render-metadata/output-model.spec.ts|features=signalInputs,outputFn
 file|tests/mock-render-metadata/output-observable.spec.ts|features=outputFn
 file|tests/mock-render-metadata/query-signals.spec.ts|features=signalQueries
--- a/tests/issue-14918/test.spec.ts
+++ b/tests/issue-14918/test.spec.ts
@@ -1,0 +1,148 @@
+import {
+  booleanAttribute,
+  Component,
+  Directive,
+  Input,
+} from '@angular/core';
+import { TestBed } from '@angular/core/testing';
+import { By } from '@angular/platform-browser';
+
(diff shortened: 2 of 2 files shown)


satanTime:

Issue #11214, opened 2025-03-14
Bug: Can't bind to 'property' of a third-party standalone component


kkoo95:
### Description of the bug

I have a standalone component which imports a standalone component from a third-party lib. And when i run my tests fairly simply with MockBuilder, i get warnings/errors in the console like:

```
NG0303: Can't bind to 'property' since it isn't a known property of 'ext-component' (used in the 'MyTestedComponent' component template).
```

### An example of the bug

```ts
import { ExtComponent } from '@org/lib';
import { Component } from '@angular/core';

@Component({
    selector: 'my-component',
    standalone: true,
    imports: [ExtComponent],
    template: `<ext-component [property]="42" /> `
})
export class MyTestedComponent {}
```
```ts
describe('MyTestedComponent', () => {
    beforeEach(() => MockBuilder(MyTestedComponent));

    it('should create', () => {
        expect(MockRender(MyTestedComponent)).toBeDefined();
    });
});
```

[Link](https://stackblitz.com/edit/github-fxmu3axs?file=package.json,src%2Ftest.spec.ts)

### Expected vs actual behavior

I shouldnt give an error

iamkinetic:
This is a #9698 duplicate.

satanTime:
I could not reproduce this on current `main` with a focused standalone repro in the repo, so this is likely already fixed by the work that landed for #9698 via PR #11909.

Please verify this against the latest `ng-mocks` release (`v14.15.2` at the time of closing). If you can still reproduce it there, feel free to reopen the issue with a minimal reproduction.

Issue #9698, opened 2024-08-13
Bug: InputSignal not working when using NgMocks on a component imported from an external library


iamkinetic:
### Description of the bug

MockComponent on component imported from an external library does not work with InputSignal.

### An example of the bug

This is the simplified imported component definition:
```typescript
export declare class SomeComponent {
    isDisabled: InputSignal<boolean>;
    isInProgress: boolean;
}
```
Used like this: 
```html
 <some-component [isDisabled]="true" [isInProgress]="true" />
```

The `isInProgress` is using the `@Input()` decorator and is correctly working/rendering.
The isDisabled is a signal input and doesn't seem to be correctly rendered/working.

```
 Error: NG0303: Can't bind to 'isDisabled' since it isn't a known property of 'some-component' (used in the 'TestComponent' component template).
        1. If 'some-component' is an Angular component and it has the 'isDisabled' input, then verify that it is included in the '@Component.imports' of this component.
        2. If 'some-component' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@Component.schemas' of this component to suppress this message.
        3. To allow any property add 'NO_ERRORS_SCHEMA' to the '@Component.schemas' of this component.
        error properties: Object({ code: 303 })
            at reportUnknownPropertyError (node_modules/@angular/core/fesm2022/core.mjs:10087:15)
            at handleUnknownPropertyError (node_modules/@angular/core/fesm2022/core.mjs:10083:5)
            at elementPropertyInternal (node_modules/@angular/core/fesm2022/core.mjs:12061:17)
            at ɵɵproperty (node_modules/@angular/core/fesm2022/core.mjs:22458:9)
            at templateFn (ng:///CitizenPortalSelectionComponent.js:25:9)
            at executeTemplate (node_modules/@angular/core/fesm2022/core.mjs:11620:9)
            at refreshView (node_modules/@angular/core/fesm2022/core.mjs:13238:13)
            at detectChangesInView (node_modules/@angular/core/fesm2022/core.mjs:13454:9)
            at detectChangesInViewIfAttached (node_modules/@angular/core/fesm2022/core.mjs:13414:5)
```

If I remove the [IsDisabled]="true" from the html, the page renders correctly and the unit test are ok.

If I copy the component from the library directly in my angular project, the InputSignal is correctly working with MockComponent.  Is something missing from the external library or is MockComponent not able to read the InputSignal from an exported component?

[...]

PR #14952, opened 2026-09-12
test(metadata): cover compiled bindings and align integration lint #14919

### LINKED ISSUE #14919: Cover precompiled library signal inputs outputs and model bindings
## Context

Follow-up to #9698, #11214 and #11909.

The original report distinguished signal bindings on components from an external compiled library from otherwise identical local declarations. Related reports also identified outputs and model bindings on that metadata surface.

## Coverage to add

Current [metadata collection](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/libs/ng-mocks/src/lib/resolve/collect-declarations.ts#L333), synthetic reflection tests and local compiled regressions cover many input/output variants. A focused integration regression should also establish behavior when the dependency comes from a compiled library and its signal metadata must be recovered from Angular definitions rather than local property decorators.

- Use an actual precompiled dependency component and directive through the existing integration/build surfaces. Ensure the fixture retains the metadata shape that distinguishes the external-library report from a local source declaration.
- Cover aliased and unaliased signal inputs, an output, and a model input/output roundtrip. For mocked models, update the parent through the separate emitter returned by `ngMocks.output`; do not assume the mock input signal implements `model.set` or `model.update`. Include required inputs where supported by the chosen fixture/version.
- Exercise direct dependency mocking and MockBuilder/module consumption; assert signal values after repeated parent updates, public input/output lookup, emitted values and model signal identity.
- Keep an equivalent local declaration and an ordinary decorator binding as controls. Preserve the documented transform distinction: mocked signal inputs receive raw bound values because their initializer transforms are omitted, while reflected decorator transforms remain effective. Assert mock identity so accidentally keeping the real dependency cannot make the test pass.
- Gate each API at its actual introduction version and establish the cases in compiled runners.

This is an external-library coverage gap, not a claim that the current implementation fails these cases. Investigate and fix any remaining shared metadata defect exposed by the regressions.

satanTime:
## Why

#9698 and #11214 reported signal bindings that differed between external compiled dependencies and local declarations. Local and synthetic metadata coverage did not establish that the compiled-library input, output, and model paths work together.

Integration lint also depended on local dependency installs. The commit hook ran, but resolved types made ES-X treat imported methods differently from CI, which lints before installing integration dependencies.

## What

Add integration regressions using the published Angular Material timepicker component and input directive, both through direct mocks and MockBuilder module consumption. Keep normal package metadata as a control and isolate compiled-definition fallback by temporarily hiding development property annotations and clearing reflection caches. Preserve the actual compiled definitions and restore the original descriptors after each case.

Cover aliased, unaliased, and required signal inputs, repeated parent writes, output delivery, public binding lookup, stable signal identity, and model round trips through the separate mock output emitter. Equivalent local declarations and a decorated Material input distinguish omitted signal initializer transforms from effective decorator transforms.

Use syntax-based lint analysis for the integration project so optional dependency installs cannot alter ES-X inference. Keep the existing narrow explanations for `ngMocks.find` calls, which are not array methods.

## Impact

The existing implementation handles these cases; the regressions run in the Angular 22 integration project against its precompiled dependencies. Local hooks and CI apply the same ES compatibility checks to these integration calls. Root typed lint and all enabled ES restrictions remain in place. Runtime behavior, public APIs, and public documentation do not change.

Fixes #14919

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -292,6 +292,15 @@ export default defineConfig([
     rules: tsEslintPlugin.configs['eslint-recommended'].overrides[0].rules,
   },
   {
+    files: ['tests-e2e/**/*.ts'],
+    languageOptions: {
+      parserOptions: {
+        // CI lints before these dependencies are installed; local installs must not change the rules.
+        project: false,
+      },
+    },
+  },
+  {
     files: ['**/*.spec.ts'],
     rules: {
       // Test fixtures may intentionally exercise eager change detection.
--- a/tests-e2e/src/issue-14919/local.spec.ts
+++ b/tests-e2e/src/issue-14919/local.spec.ts
@@ -103,11 +103,13 @@ describe('issue-14919:local', () => {
         const fixture = TestBed.createComponent(HostComponent);
         fixture.detectChanges();
         const host = fixture.componentInstance;
+        // eslint-disable-next-line es-x/no-array-prototype-find -- ngMocks.find is not Array.find.
         const componentElement = ngMocks.find(
           fixture,
           LocalTimepickerComponent,
         );
         const component = componentElement.componentInstance;
+        // eslint-disable-next-line es-x/no-array-prototype-find -- ngMocks.find is not Array.find.
         const directiveElement = ngMocks.find(fixture, 'input');
         const directive = ngMocks.get(
           directiveElement,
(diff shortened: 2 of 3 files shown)


satanTime:

Issue #10960, opened 2025-02-05
Bug: Importing ReactiveFormsModule in standalone ControlValueAccessor


menrodriguez-cc:
### Importing ReactiveFormsModule in standalone ControlValueAccessor throws "TypeError: Cannot read properties of undefined (reading 'constructor')"

First of all, excuse me if this is not a bug but a misuse, as I am trying ng-mocks for the first time.

I am trying to use a standalone ControlValueAccessor as a nested form. I know there are many opinions against this use of CVAs, but I have used this approach in the past and it works for me. 

I have followed the example for [testing form controls](https://ng-mocks.sudo.eu/extra/mock-form-controls) and the one at issue [3778](https://github.com/help-me-mom/ng-mocks/issues/3778). 

If I keep the "imports: [ReactiveFormsModule]" in the standalone ControlValueAccessor error `TypeError: Cannot read properties of undefined (reading 'constructor')` is thrown

But I think this import is necessary, as directive  "formControlName" is used in the component's template. 

### An example of the bug

Link: https://stackblitz.com/edit/github-ddxsqzb1?file=src%2Ftest.spec.ts

### Expected vs actual behavior

A standalone ControlValueAccessor should be able to import ReactiveFormsModule and have this dependency kept with MockBuilder.

satanTime:
[v14.15.2](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.15.2) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14954, opened 2026-09-12
test(MockBuilder): cover real kept standalone value accessors #14922

### LINKED ISSUE #14922: Cover real kept standalone value accessors importing ReactiveFormsModule
## Context

Follow-up to [#10960](https://github.com/help-me-mom/ng-mocks/issues/10960) and [#13305](https://github.com/help-me-mom/ng-mocks/pull/13305).

The fix preserves `useExisting` aliases while processing kept providers, including the `NG_VALUE_ACCESSOR` self alias of a standalone control that imports `ReactiveFormsModule`. The retained regression does not establish the real kept-control behavior described by its comments.

## Coverage gap

The [current setup](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/tests/issue-10960/test.spec.ts#L99) uses `MockBuilder(TargetComponent).keep(ReactiveFormsModule)`. The standalone host's dependencies are shallow-mocked, so this does not explicitly keep `StandaloneCVAComponent`. The test customizes `writeValue` through `MockInstance` and drives `ngMocks.change` on the accessor; it never proves that the original accessor or its inner form control is running.

The relevant [provider resolution branch](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/libs/ng-mocks/src/lib/mock-service/helper.resolve-provider.ts#L165) distinguishes kept targets, `skipMock`, and explicit provider decisions. Existing generic alias and CVA coverage does not retain this exact real standalone control/import combination.

This is a coverage gap, not a claim of a current runtime failure.

## Acceptance

- Preserve the current mocked-dependency case and add an independent setup that explicitly keeps `StandaloneCVAComponent` as well as `ReactiveFormsModule`.
- Assert that the accessor is real, and that its element injector's `NG_VALUE_ACCESSOR` array contains exactly that same accessor instance.
- Verify that the real inner input/control receives the initial host value; an inner-control change reaches the parent once; a parent change reaches the inner input; and disable/enable state propagates correctly.
- Keep these behavioral assertions independent of a mocked `writeValue` implementation, so a shallow-mocked accessor cannot make them pass.
- Retain explicit provider override policy, using the existing alias/CVA controls or a focused source unit if the selected branch lacks direct coverage. Correct the misleading retained-test comments as appropriate.
- Investigate and fix only a shared defect demonstrated by the new regressions.

The [#3791 suite](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/tests/issue-3791/test.spec.ts) already covers a kept CVA and explicit forms-token decisions, but that CVA does not itself import `ReactiveFormsModule`. The new case should preserve the distinguishing #10960 setup rather than expand into a generic forms matrix.

satanTime:
## Why

The regression for #10960 shallow-mocks the standalone value accessor, despite comments claiming that it keeps the real control. It covers the mock's forms handshake but does not establish that a kept accessor can import `ReactiveFormsModule` and retain its self-referencing provider.

## What

Add an independent regression that explicitly keeps the accessor and `ReactiveFormsModule`. It verifies that the element injector provides exactly the real accessor instance, the inner input receives the initial value, an inner change reaches the parent once, parent writes reach the input, and disable/enable state propagates.

Preserve the original mocked case, assert its mock identity, and correct its comments. Existing explicit forms-token override controls remain applicable.

## Impact

Both the real and mocked standalone-control paths have functional regression coverage. The current implementation already supports the real case; no runtime or public documentation change is needed.

Closes #14922

### DIFF
--- a/test-spread.conf
+++ b/test-spread.conf
@@ -140,6 +140,7 @@ file|tests/mock-render-metadata/static-queries.spec.ts|versions=8+
 file|tests/mock-render-metadata/input-transforms.spec.ts|versions=16+
 file|tests/issue-14918/test.spec.ts|versions=16+
 file|tests/issue-14920/test.spec.ts|features=signalQueries,hostDirectives
+file|tests/issue-14922/test.spec.ts|features=standalone
 file|tests/mock-render-metadata/output-model.spec.ts|features=signalInputs,outputFn
 file|tests/mock-render-metadata/output-observable.spec.ts|features=outputFn
 file|tests/mock-render-metadata/query-signals.spec.ts|features=signalQueries
--- a/tests/issue-10960/test.spec.ts
+++ b/tests/issue-10960/test.spec.ts
@@ -8,23 +8,14 @@ import {
 } from '@angular/forms';
 
 import {
+  isMockOf,
   MockBuilder,
   MockInstance,
   MockRender,
   ngMocks,
 } from 'ng-mocks';
 
-// This standalone CVA is the minimal reproduction of the reported issue:
-// - it imports ReactiveFormsModule by itself
-// - it provides NG_VALUE_ACCESSOR via useExisting + forwardRef to itself
-//
-// Before the fix, MockBuilder kept this declaration in skipMock mode, but the
-// provider override logic still rewrote the useExisting target. That broke
-// Angular's value-accessor lookup and rendering the host form crashed with:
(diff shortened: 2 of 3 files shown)


satanTime:

PR #14955, opened 2026-09-12
fix(ngMocks): resolve input queries against their declaring directives #14923

### LINKED ISSUE #14923: Check attribute-value lookup across directives sharing an input alias
## Context

Follow-up to #11182 and #11179, which changed the shared attribute/value lookup used by `ngMocks.reveal` and template-reference helpers.

## Suspected wider defect

[The Ivy crawler](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/libs/ng-mocks/src/lib/mock-helper/crawl/crawl-by-attribute-value.ts#L38) iterates the directive indices associated with an input. Each iteration then calls `mockHelperAttributes` for the node and public alias. [That helper](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/libs/ng-mocks/src/lib/mock-helper/mock-helper.attributes.ts#L35) returns the first matching provider's value, independently of the current directive index.

When two directives on one node expose the same public input alias but hold different values, a value held by the second directive may never be considered. Existing reveal tests have shared aliases, but identical values do not distinguish the providers. This is a source-supported suspicion that still needs a failing runtime reproduction.

- Put two real/kept directives with the same public input alias on an `ng-template`. Make their stored values differ through input normalization or an instance update, and reverse the declaration order. Add an `ng-container` variant for the reveal helpers where useful; template-reference helpers require an actual template.
- Query each value through `reveal`/`revealAll` and `findTemplateRef`/`findTemplateRefs`, asserting the expected node or template identity and no false matches.
- Check aliased signal/transformed inputs on real/kept directives where they use the same crawler, with the appropriate compiler/version gates. Mock signal initializer transforms are intentionally omitted and must not be relied on to create differing values.
- Preserve ordinary aliases, single-directive matches, static/dynamic bindings and not-found behavior.
- Pair a real-Angular reproducer with focused source-unit coverage of multiple directive indices. Fix the shared value-resolution path if the suspected mismatch is confirmed.

satanTime:
## Why

When two directives on one node expose the same public input alias but store different values, the shared Ivy crawler repeatedly reads the first provider. `ngMocks.reveal`, `revealAll`, `findTemplateRef`, and `findTemplateRefs` can therefore miss a value held by another directive. This follows the lookup fixes for #11179 and #11182.

View Engine can also share a parent's provider map with its children, causing input queries to return descendants that do not declare the matching input.

## What

Resolve input metadata and values from each indexed directive instance. Read stored transformed values and unwrap signal inputs without applying transforms again. Exclude numeric input flags from directive lookup so they cannot produce a false match for `undefined`.

Include the separate host-directive input map used by newer Angular versions, resolving each entry through its original input alias.

Restrict View Engine provider lookup to the declaring node, preserving each child's own inputs for both presence and value queries.

Add real-Angular regressions for both directive orders, static and dynamic template bindings, containers, decorator transforms, signal inputs, host directives, updates, missing values, and duplicate-free results. Focused source tests cover the shared-alias lookup and preserve callable inputs, signal identity, host alias mappings, and metadata fallbacks.

## Impact

All four lookup helpers can find either directive's stored value through the shared alias without falsely matching inherited inputs. Ordinary `ngMocks.input` first-provider behavior remains unchanged. The fix restores existing functionality without new public APIs or documentation changes.

Fixes #14923


satanTime:

PR #13312, opened 2026-03-15
#7937 fix: allow TestBed.inject overrides for mocked declarations

satanTime:
Closes #7937

## Summary
- track explicit TestBed.inject/get calls for mocked declarations
- replay seeded descriptor overrides onto later declaration-local mock instances
- add regression coverage for pipes, directives, and components plus internal helpers/overrides

## Validation
- COMPOSE_PROJECT_NAME=ngmocks_7937 sh ./test.sh root
- COMPOSE_PROJECT_NAME=ngmocks_7937 sh ./test.sh coverage

Issue #7937, opened 2024-01-13
Bug: cannot set return value of DI'ed pipe


JPtenBerge:
HI there!

### Description of the bug

I try to set the return value of a DI'ed pipe's `transform()` function, but it doesn't. I also `expect()` the function `.toHaveBeenCalled()`, but it's never called. It's as if my test and my component have different instances of the pipe.

It could simply be that I'm doing something wrong or just forgot something somewhere. Couldn't find this specific use case in the docs and I believe I'm doing everything correctly, so I'd thought I'd file a bug.

### An example of the bug

I have this relatively boring pipe:

```ts
@Pipe({ name: 'bla' })
export class BlaPipe implements PipeTransform {
  transform(value: any, ...args: any[]) {
    return `hi there ${value}`;
  }
}
```
It resides in this module, where I `provide` the pipe for use with DI:
```ts
@NgModule({
  declarations: [BlaPipe],
  exports: [BlaPipe],
  providers: [BlaPipe],
})
export class PipesModule {}
```
I then try to use this pipe:
```ts
@Component({
  selector: 'app-root',
  standalone: true,
  imports: [PipesModule],
  templateUrl: './app.component.html',
})
export class AppComponent {
  constructor(private blaPipe: BlaPipe) {}

  doSomething() {
    let val = this.blaPipe.transform('Frank');
    console.log('result of pipe transform:', val);
    return val;
  }
}
```
Which all works as expected. But then I try to test it. I have the `autoSpy()` set:
```ts
ngMocks.autoSpy('jasmine');

getTestBed().initTestEnvironment(BrowserDynamicTestingModule, platformBrowserDynamicTesting());
```
and then the test itself:
```ts
describe('AppComponent', () => {
  let blaPipe: jasmine.SpyObj<BlaPipe>;
  let sut: AppComponent;

  beforeEach(() => MockBuilder(AppComponent));

  beforeEach(() => {
    blaPipe = TestBed.inject(BlaPipe) as jasmine.SpyObj<BlaPipe>;
    // blaPipe = ngMocks.findInstance(BlaPipe) as jasmine.SpyObj<BlaPipe>;

    blaPipe.transform.and.returnValue('test');

    sut = TestBed.createComponent(AppComponent).componentInstance;
  });

  it('should use the blaPipe', () => {
    let result = sut.doSomething();

    expect(blaPipe.transform).toHaveBeenCalled();
    expect(result).toBe('test');
  });
});
```
[...]

PR #14957, opened 2026-09-12
fix(ngMocks): isolate replayed declaration accessor values #14924

### LINKED ISSUE #14924: Check per-instance accessor state when replaying TestBed declaration seeds
## Background

Follow-up to [#7937](https://github.com/help-me-mom/ng-mocks/issues/7937) and its [shared declaration-provider fix](https://github.com/help-me-mom/ng-mocks/commit/d0743f6369d17f5c8c7255ee25a1880ad4bf9553), also associated with #13312.

The fix replays an explicitly injected TestBed seed onto a declaration instance that Angular creates later. This preserves configured method/spy identities, as covered for a [pipe](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-7937/test.spec.ts#L59) and a [component](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-2646/test.spec.ts#L260).

An adjacent risk affects ordinary getter/setter inputs, even when the test never customizes that input. The root declaration provider creates a [MockService instance](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/mock-provider.ts#L29). [MockService creates own mock accessors from the original prototype](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/helper.create-mock-from-prototype.ts#L18). Their [setter and getter are connected](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/helper.mock.ts#L63) through [closure-backed value storage](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/helper.mock-function.ts#L19).

Each rendered declaration initially receives its own mock accessors in [the mock constructor](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/common/mock.ts#L250). However, [seed replay](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/helper.replay-instance.ts#L17) copies the seed's accessor descriptor, including the same getter/setter functions, into each later instance. That appears to make separate input bindings share one closure-backed value. This is a source-supported risk; the two-instance runtime behavior still needs a focused reproducer.

The normal MockService seed does not initialize original signal-input or output fields. Those fields are not the basis of this report.

## Expected preservation

Explicit seed method overrides should continue to reach rendered declarations with the intended spy identity. Unmodified generated accessor backing state should remain local to each declaration instance.

- Start from the existing `ServiceToComponent extends TargetService` provider setup, where `TargetService` is injectable. Add an ordinary `@Input` getter/setter on the component prototype, inject the mock before rendering, and customize only its `echo` method.
- Render two instances of that same mock component with independently bound values `A` and `B`. Assert each component and `ngMocks.input` reports its own value, then update only one binding and confirm the other remains unchanged.
- Include an ordinary field input as a control: its separately bound per-instance values must remain independent as well.
- Assert both instances retain the configured method/spy. Do not run the real getter/setter or constructor merely to maintain independent mock input values.
- Include a no-seed control and exercise a mocked attribute directive through the same shared accessor/replay path where practical. Preserve the existing pipe seed regression and reset behavior.
- Preserve deliberate accessor stubs on the seed. Define that customization behavior explicitly if the fix must distinguish a generated accessor from one the user replaced or configured.
- If the registry's single `resolved` pointer needs adjustment, first establish the intended `TestBed.inject` behavior for multiple local instances; do not silently introduce a new injector-identity contract.

Use a focused source-unit regression alongside real Angular bindings, and change replay behavior only for the demonstrated accessor state-sharing problem. This is separate from #2713 / #14818, which concern applying later `MockInstance` registrations to already existing instances.

satanTime:
## Why

When a mock declaration was injected before rendering, seed replay copied its generated getter and setter functions into every later instance. Those functions shared one backing value, so two independently bound component or directive inputs could overwrite each other and the injected seed.

This follows the declaration-provider seed replay introduced for #7937 and #13312.

## What

Replay generated accessor pairs with independent storage initialized from the seed's current value. Continue copying deliberately customized accessors and method spies with their original identities.

Add real Angular regressions for two instances of both a component and an attribute directive, including independent binding updates, ordinary field inputs, no-seed behavior, and explicit accessor stubs. Source tests cover configured seed values, later instances, descriptor preservation, and existing target accessors without invoking them.

## Impact

Tests can configure a declaration's methods before rendering while keeping each rendered instance's input values independent. The existing injector identity and reset behavior is unchanged. Public APIs and documentation are unchanged.

Fixes #14924

### DIFF
--- a/libs/ng-mocks/src/lib/mock-service/helper.replay-instance.spec.ts
+++ b/libs/ng-mocks/src/lib/mock-service/helper.replay-instance.spec.ts
@@ -1,4 +1,8 @@
+import { ngMocks } from '../mock-helper/mock-helper';
+
+import helperDefinePropertyDescriptor from './helper.define-property-descriptor';
 import helperReplayInstance from './helper.replay-instance';
+import { MockService } from './mock-service';
 
 describe('helper.replay-instance', () => {
   it('returns early for empty or identical instances', () => {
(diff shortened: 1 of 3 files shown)


satanTime:

PR #7442, opened 2023-11-15
fix(ngMocks): respects custom errors on lookups #7041

satanTime:
closes #7041

Issue #7041, opened 2023-10-20
Bug: Angular 16 ngMocks.findInstance swallows original error


CSchulz:
### Description of the bug

If there is an error with the provider factory during the tests ngMocks.findInstance swallows the original error and shows `Cannot find an instance via ngMocks.findInstance(...)` only.
This requires debugging into ng-mocks to find the root cause.

### An example of the bug

https://stackblitz.com/edit/github-6z37vt?file=src%2Ftest.spec.ts

### Expected vs actual behavior

Expected: Show also the original error if there is one.

`Cannot find an instance via ngMocks.findInstance(...). Caught an error during the injection call: <original error>`.

Actual: `Cannot find an instance via ngMocks.findInstance(...)`

satanTime:
Hi @CSchulz,

thanks for reporting the issue. It will be fixed with the next release.

satanTime:
[v14.12.0](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.12.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

Issue #2398, opened 2022-05-07
Bug: MockRender and pipes


satanTime:
```ts
import {MockBuilder, MockRender, ngMocks} from 'ng-mocks';

import { Pipe, PipeTransform } from '@angular/core';

@Pipe({
  name: 'phone'
})
export class PhonePipe implements PipeTransform {

  transform(value: any, args?: any): any {
    const inputVal = value.toString();
    const slice1 = inputVal.slice(0, 3);
    const slice2 = inputVal.slice(3, 6);
    const slice3 = inputVal.slice(6)
    return `+1(${slice1})-${slice2}-${slice3}`;
  }
}

describe('Phone Pipe', () => {
  beforeEach(() => {
    return MockBuilder(PhonePipe).provide(PhonePipe)
  })
  it('should be defined', () => {
    const fixture = MockRender(PhonePipe);
    expect(fixture.point.componentInstance).toBeDefined();
  })
  it('should test the transform value', () => {
    const fixture = MockRender('{{ "4161234567" | phone }}');
    expect(ngMocks.formatText(fixture)).toBe('+1(416)-123-4567');
  })
})
```

satanTime:
[v13.5.1](https://github.com/ike18t/ng-mocks/releases/tag/v13.5.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14958, opened 2026-09-12
fix(ngMocks): preserve provider errors across lookup and pipe rendering #14926

### LINKED ISSUE #14926: Check provider error propagation across lookup and pipe-rendering paths
## Background

Follow-up to [#7041](https://github.com/help-me-mom/ng-mocks/issues/7041), fixed by [PR #7442](https://github.com/help-me-mom/ng-mocks/pull/7442) / [6c78f548a](https://github.com/help-me-mom/ng-mocks/commit/6c78f548a), and the provided-pipe rendering behavior from [#2398](https://github.com/help-me-mom/ng-mocks/issues/2398).

#7041 preserves provider factory errors for `ngMocks.findInstance(Token)` before a fixture exists and for one-argument `ngMocks.get(Token)`. The [existing regression](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-7041/test.spec.ts#L59) covers those routes, including nested service dependencies and genuinely absent providers.

Adjacent public routes still discard or replace unexpected injection errors:

- With no fixture, [`findInstances`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-helper/find-instance/mock-helper.find-instances.ts#L41-L46) catches every `getInjection` error and returns an empty result.
- Fixture-scoped `findInstance`, `findInstances`, and `get` use [`func.get-from-node-injector`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-helper/func.get-from-node-injector.ts#L27-L29). Its [`coreInjector` helper](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/common/core.injector.ts#L11-L17) catches every error, including failures in a present lazy provider's factory, and returns `undefined`.
- [`MockRender`'s injection route](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-render/mock-render-factory.ts#L49-L64) replaces any error for a pipe with the missing `$implicit` / provider hint. A pipe that is provided but whose constructor or provider factory throws can therefore lose its actual failure.

These are source-supported diagnostic risks; the additional runtime failures have not been reproduced during this read-only audit.

## Bounded investigation and acceptance

1. Reuse the #7041 throwing factory with no fixture and call `ngMocks.findInstances(Token)`. Require the original error to remain visible instead of receiving `[]`.
2. Render a kept component or directive with a lazy local token provider whose factory throws a distinctive error. Do not inject the token during component construction. Look it up through fixture-scoped `findInstance`, `findInstances`, and element-scoped `get`, each in a fresh setup. Require the provider error rather than a missing-instance message, fallback value, or empty collection. Include a supplied not-found fallback to distinguish absence from a broken provider.
3. Provide a real pipe with a throwing constructor or `useFactory`, then call `MockRender(Pipe)` without template params. Require the actual error rather than advice to add `$implicit`. Preserve the [#2398 missing-provider hint and successful provided-pipe cases](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-2398/test.spec.ts).
4. Preserve genuinely missing-token behavior: singular helpers keep their documented failure/fallback behavior, plural helpers return an empty array, and missing pipe injection keeps its actionable hint. Preserve ordinary successful lookup identity and deduplication.
5. Pair focused source-unit error classification/forwarding coverage with real Angular provider regressions at the supported API boundaries. Assert the original error object where Angular forwards it unchanged, or its distinctive cause/message when Angular wraps it.

The shared `coreInjector` also serves optional crawl, template, and form-control probes. Avoid blindly changing every caller to throw: define the intended public lookup behavior and preserve those optional discovery paths. Fix only the demonstrated diagnostic gaps.

satanTime:
## Why

A present provider's factory or dependency error can be replaced by a missing-instance message, fallback value, or empty collection. Rendering a provided pipe can similarly replace its real failure with advice to provide the pipe.

Fixes #14926. Follow-up to #7041 / #7442 and the provided-pipe behavior from #2398.

## What

Use Angular's native missing-provider fallback to distinguish absence from creation failures in lookup and pipe-rendering paths. Resolve requested node providers before optional parent probes can consume their original error, while retaining optional discovery of nonlocal pipes that require a different view context.

Add regressions for component and directive local providers, root and nested dependencies, pipe factories and constructors, explicit fallbacks, and successful lookup identity and deduplication.

## Impact

Provider failures retain their actionable diagnostics. Genuine absence keeps its existing behavior, and fixture searches continue to discover valid component-local pipes. The fix restores existing behavior without changing public APIs or documentation.

### DIFF
--- a/libs/ng-mocks/src/lib/mock-helper/find-instance/mock-helper.find-instance.ts
+++ b/libs/ng-mocks/src/lib/mock-helper/find-instance/mock-helper.find-instance.ts
@@ -1,4 +1,5 @@
-import { getInjection } from '../../common/core.helpers';
+import { getTestBed } from '@angular/core/testing';
+
 import { Type } from '../../common/core.types';
 import { getSourceOfMock } from '../../common/func.get-source-of-mock';
 import { isNgDef } from '../../common/func.is-ng-def';
@@ -36,13 +37,15 @@ export default <T>(...args: any[]): T => {
       true,
     );
   } else {
-    try {
-      result.push(getInjection(declaration));
-    } catch (error) {
-      // forwarding unexpected errors: https://github.com/help-me-mom/ng-mocks/issues/7041
-      if (!error || typeof error !== 'object' || (error as any).ngTokenPath === undefined) {
-        throw error;
-      }
+    const testBed = getTestBed();
+    const instance = testBed.inject
+      ? testBed.inject(declaration, defaultNotFoundValue)
+      : /* istanbul ignore next */ (testBed as typeof testBed & { get: typeof testBed.inject }).get(
+          declaration,
+          defaultNotFoundValue,
+        );
+    if (instance !== defaultNotFoundValue) {
+      result.push(instance);
     }
   }
 
--- a/libs/ng-mocks/src/lib/mock-helper/find-instance/mock-helper.find-instances.ts
+++ b/libs/ng-mocks/src/lib/mock-helper/find-instance/mock-helper.find-instances.ts
@@ -1,4 +1,5 @@
-import { getInjection } from '../../common/core.helpers';
+import { getTestBed } from '@angular/core/testing';
+
 import { Type } from '../../common/core.types';
 import { getSourceOfMock } from '../../common/func.get-source-of-mock';
 import { isNgDef } from '../../common/func.is-ng-def';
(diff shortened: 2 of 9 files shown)


satanTime:

Issue #5262, opened 2023-03-22
Bug: Maximum call stack size exceeded


lneumeier:
After the update from 14.5.0 to 14.5.1 (#4564) I'm getting following error

```
Maximum call stack size exceeded
        at Object.isExcludedDef (node_modules/ng-mocks/index.mjs:1:6637)
        at node_modules/ng-mocks/index.mjs:1:10494
        at t (node_modules/ng-mocks/index.mjs:1:10437)
        at node_modules/ng-mocks/index.mjs:1:10528
        at t (node_modules/ng-mocks/index.mjs:1:10437)
        at node_modules/ng-mocks/index.mjs:1:10528
        at t (node_modules/ng-mocks/index.mjs:1:10437)
        at node_modules/ng-mocks/index.mjs:1:10528
        at t (node_modules/ng-mocks/index.mjs:1:10437)
        at node_modules/ng-mocks/index.mjs:1:10528
```

on some of my unittests. Do you have any idea what I should look for?

PS:

I'm using 
```
declarations: [MockComponents(Example1Component, Example2Component)],
```
and 
```
declarations: [MockComponent(ExampleComponent)],
````

satanTime:
[v14.7.2](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.7.2) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

lneumeier:
Thank you for the quick fix.

satanTime:
Not at all, please let me know if it solves your issue.

lneumeier:
Yes, the issue is solved. Thank you.

satanTime:
Cool, thanks! Stay in touch!

PR #14959, opened 2026-09-12
fix(MockBuilder): preserve replacements across shared and cyclic values #14927

### LINKED ISSUE #14927: Check replacement consistency across shared and cyclic provider values
Follow-up to [#5262](https://github.com/help-me-mom/ng-mocks/issues/5262), also resolving [#5240](https://github.com/help-me-mom/ng-mocks/issues/5240), and the [recursive-mock fix](https://github.com/help-me-mom/ng-mocks/commit/c699d57fc77fc878986cddf33c18b4ac43539b3f).

## Suspected wider defect

The fix prevents infinite recursion, but repeated references can escape the replacement/exclusion applied to their first occurrence. The provider-value walker [caches the newly constructed array/object](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/helper.replace-with-mocks.ts#L22), then [returns the original value on a cache hit](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/helper.replace-with-mocks.ts#L101).

There is an ordinary public route: keep a `useValue` configuration token from a mocked module, and exclude a declaration referenced by that configuration. [Provider resolution walks kept values](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/helper.resolve-provider.ts#L84), and the existing [object-exclusion unit](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/mock-service.spec.ts#L440) already establishes this policy. If the value contains `{ first: shared, second: shared }` and `shared` references the excluded declaration, the first path is rebuilt without that declaration while the second path returns the original `shared` object. A back-edge to a changed object has the same risk.

This is a source-supported likely defect; a focused public regression should establish the observed behavior before changing production code. The [retained issue regression](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-5262/test.spec.ts#L28) checks that recursive values do not throw, but does not inspect replacement consistency or graph identity.

- Start with an acyclic shared object in a module's `useValue` provider. Use `MockBuilder().mock(TargetModule).keep(CONFIG).exclude(ExcludedDeclaration)`. Assert the excluded reference is absent through both paths, and the original supplied configuration remains unchanged.
- Include a replaced declaration or mocked declaration reference through the same shared-node path. Assert that both paths resolve to the intended replacement/mock identity.
- Extend the same value to a self-reference or mutual reference with an actual replacement/exclusion. Assert traversal terminates and references within the returned changed graph cannot lead back to an unprocessed original node. Keep repeated references consistent within that result.
- Preserve identity for a wholly unchanged value graph; the existing walker intentionally returns the original value when no changes are needed. Do not turn this into unconditional deep cloning.
- Cover arrays and objects through focused source units, including the route guard/resolver filtering branches if the cache result must account for the additional objects those branches create. Retain the existing recursive public regression and compatibility boundaries.

This concerns successful value-graph traversal. It is separate from [#14911](https://github.com/help-me-mom/ng-mocks/issues/14911), which concerns resolver cleanup after failed declaration construction.

satanTime:
## Why

The recursion fix in #5262 stopped infinite traversal, but later references to a shared or cyclic provider value could still expose the original declarations. Explicit `.replace()` choices were also missed because the walker only consulted the declaration mock cache.

## What

Keep one processed object or array for each changed value and propagate changes through its references. Resolve unchanged graph regions back to their original objects. Read explicit replacement decisions through the existing policy helpers and preserve dependency token identity during guard and resolver filtering.

Add regressions for shared objects and arrays, self and mutual references, exclusions, replacements, mock identity, unchanged values, and route filtering with shared aliases.

## Impact

Kept configuration providers consistently contain the selected declarations through every reference, while the supplied configuration remains intact. This restores expected behavior and requires no public API or documentation changes.

Fixes #14927.

### DIFF
--- a/libs/ng-mocks/src/lib/mock-service/helper.replace-with-mocks.spec.ts
+++ b/libs/ng-mocks/src/lib/mock-service/helper.replace-with-mocks.spec.ts
@@ -1,0 +1,344 @@
+import { InjectionToken } from '@angular/core';
+
+import {
+  NG_MOCKS_GUARDS,
+  NG_MOCKS_RESOLVERS,
+} from '../common/core.tokens';
+import ngMocksUniverse from '../common/ng-mocks-universe';
+import { MockBuilderStash } from '../mock-builder/mock-builder-stash';
+
+import helperReplaceWithMocks from './helper.replace-with-mocks';
+
+describe('helper.replace-with-mocks', () => {
+  const stash = new MockBuilderStash();
+
+  beforeEach(() => stash.backup());
+  afterEach(() => stash.restore());
+
+  it('uses one processed object for every shared reference', () => {
+    class Original {}
+    class Replacement {}
+    class Excluded {}
+
+    ngMocksUniverse.cacheDeclarations.set(Original, Replacement);
+    ngMocksUniverse.builtDeclarations.set(Excluded, null);
+    const shared = Object.freeze({
+      declaration: Original,
+      excluded: Excluded,
+    });
+    const value = Object.freeze({ first: shared, second: shared });
+
+    const actual: typeof value = helperReplaceWithMocks(value);
+
+    expect(actual).not.toBe(value);
+    expect(actual.first).not.toBe(shared);
+    expect(actual.first).toBe(actual.second);
+    expect(actual.first.declaration).toBe(Replacement);
+    expect(actual.second.declaration).toBe(Replacement);
+    expect('excluded' in actual.first).toBe(false);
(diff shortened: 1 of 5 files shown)


satanTime:

PR #14960, opened 2026-09-12
fix(MockService): cache function mocks without invoking them #14928

### LINKED ISSUE #14928: Check pristine call history and function identity in MockService caches
Follow-up to [#5262](https://github.com/help-me-mom/ng-mocks/issues/5262) and its [per-call mock cache change](https://github.com/help-me-mom/ng-mocks/commit/c699d57fc77fc878986cddf33c18b4ac43539b3f).

## Suspected wider defect

The function branch of `MockService` [creates a mock function and caches `value()`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/mock-service.ts#L23). That expression invokes the newly created mock during construction and stores its return value, rather than the function. With auto-spy enabled, this appears to record a call before the test or consumer calls the mock. Because an ordinary empty mock returns `undefined`, the [cache lookup](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/mock-service.ts#L56) also fails to reuse that function for a repeated source reference.

Directly passing a function is an existing supported test path: the [normal-function unit](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/mock-service.spec.ts#L130) checks callability and an undefined return. The [explicit object-shape unit](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-service/mock-service.spec.ts#L265) exercises nested function fields. Neither establishes a pristine call history immediately after construction.

This is a source-supported likely defect; retain a failing regression before correcting the cache entry. The original supplied function is not called by this expression—the unwanted call is to the new mock/spy.

- Enable the runner's auto-spy factory, create a direct `MockService(originalFunction)`, and assert the resulting spy has no calls immediately after construction. Call it once and assert exactly that one call, its arguments, and the normal undefined return. An original-function counter must remain zero throughout.
- Repeat through a supported explicit object shape containing an own function field, and through an explicitly supplied `MockService(shape)` provider override where useful. Do not assume that an uninitialized class mock discovers instance fields.
- Put the same original function in two own data fields of one shape. Assert both references reuse the cached mock and share only the calls made by the test. Different original functions and separate `MockService` calls should continue to receive independent mocks.
- Preserve auto-spy-default callability, custom mock-function factory behavior, prefix naming, ordinary prototype method isolation, and constructor suppression. Add focused source coverage and a public regression suitable for the existing runner matrix.

This is distinct from [#14916](https://github.com/help-me-mom/ng-mocks/issues/14916), which concerns evaluating original own getters while reading object shapes, and from automatic reactive-field discovery in [#2701](https://github.com/help-me-mom/ng-mocks/issues/2701).

satanTime:
## Why

Creating a function mock invokes it while populating the cache. This adds a call to a new spy and caches its return value, so repeated references receive separate mocks or even a configured return value instead of a function. A custom mock that throws also makes construction fail.

The same path affects function fields in explicit object shapes and mocked object providers, following the recursive-cache change in #5262.

## What

Cache the generated function itself. Add regressions for pristine call history, shared function references, independent mocks, provider overrides and automatically mocked `useValue` providers, including default and custom spy factories.

## Impact

Function mocks are called only by their consumers, and repeated references within one `MockService` operation share the same mock. Separate operations and prototype methods remain independent. Existing arguments, naming, constructor suppression, and documented behavior are preserved; no public API or documentation changes are needed.

Fixes #14928

### DIFF
--- a/libs/ng-mocks/src/lib/mock-service/mock-service.spec.ts
+++ b/libs/ng-mocks/src/lib/mock-service/mock-service.spec.ts
@@ -137,6 +137,115 @@ describe('MockService', () => {
     expect(mockService.and.identity).toBe('func:test');
   });
 
+  it('caches a custom mock function rather than its configured return value', () => {
+    let originalCalls = 0;
+    const original = (value: string) => {
+      originalCalls += 1;
+
+      return { value };
+    };
+    const result = { value: 'configured' };
+    const generated: jasmine.Spy[] = [];
+    const shape = { first: original, second: original };
+
+    ngMocks.autoSpy(name => {
+      const mock = jasmine.createSpy(name).and.returnValue(result);
+      generated.push(mock);
+
+      return mock;
+    });
+    try {
+      const mock = MockService<typeof shape>(shape, 'custom');
+
+      expect(generated.length).toBe(1);
+      expect(generated[0].and.identity).toBe('func:custom.first');
+      expect(generated[0]).not.toHaveBeenCalled();
+      expect(mock.first).toBe(generated[0]);
+      expect(mock.second).toBe(mock.first);
+      expect(originalCalls).toBe(0);
+
+      expect(mock.first('first')).toBe(result);
+      expect(mock.second('second')).toBe(result);
+      expect(generated[0]).toHaveBeenCalledTimes(2);
+      expect(generated[0].calls.allArgs()).toEqual([
+        ['first'],
+        ['second'],
+      ]);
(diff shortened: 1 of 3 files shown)


satanTime:

PR #14961, opened 2026-09-12
fix(MockBuilder): retain providers across repeated module imports #14929

### LINKED ISSUE #14929: Check forRoot provider preservation when repeated imports include a bare module
Follow-up to [#4613](https://github.com/help-me-mom/ng-mocks/issues/4613) and the [root ModuleWithProviders fix](https://github.com/help-me-mom/ng-mocks/commit/3ed8ae4d46bb5f56228175951331b2efb579a6cb), with the mixed TestBed reconstruction introduced around [#4344](https://github.com/help-me-mom/ng-mocks/issues/4344).

## Suspected wider defect

Repeated top-level imports can lose providers when the same module appears first as `ModuleWithProviders` and later as a bare module. [`keep()` and `mock()`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.ts#L138) retain the prior provider list temporarily, call `wipe()`, and restore that list only if the new input itself supplies `providers`. A [bare module normalizes to `providers: undefined`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.ts#L35), so its earlier `forRoot()` providers appear to be discarded.

This also affects a public setup that does not explicitly overwrite any builder configuration: `TestBed.configureTestingModule({ imports: [TargetModule.forRoot(), TargetModule, MockModule(UnrelatedModule)] })`. The mixed real/mock path [retains each top-level entry](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts#L257), then [calls the builder for each entry in order](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts#L312). Reversing the two real imports feeds the provider-bearing entry last instead.

This is a source-supported likely defect, not a confirmed runtime result. Establish the corresponding real Angular behavior before changing the shared accumulation policy.

- Use a real Angular baseline with `[TargetModule.forRoot(), TargetModule]`, where only `forRoot()` supplies a distinctive token or non-root service. Assert that the provider exists with its configured identity/value and that its factory runs only when expected.
- Add an unrelated explicit `MockModule` so the same TestBed configuration takes the mixed reconstruction path. Preserve the baseline provider and host behavior, and assert that the unrelated module's dependency is actually mocked.
- Reverse the two real imports. Both orders should preserve the `forRoot()` provider. Include a repeated provider-bearing entry with multi providers as a control for the existing accumulation order; do not silently duplicate provider execution beyond Angular's own behavior.
- Cover the corresponding same-mode builder chains, such as `.keep(TargetModule.forRoot()).keep(TargetModule)`, with focused source units. Include the mock path with an explicitly kept provider token if it exercises the same preservation rule. Keep deliberate transitions between keep/mock/exclude and explicit `.provide()` precedence separate from same-mode repeated imports.
- Preserve the existing [nested-provider and precedence regressions](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-4613/providers.spec.ts#L139) and the [multiple provider-bearing overload units](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.spec.ts#L140). Change production behavior only for the demonstrated loss.

This is separate from [#14917](https://github.com/help-me-mom/ng-mocks/issues/14917), which addresses recipes in root `@Injectable` metadata, and [#4701](https://github.com/help-me-mom/ng-mocks/issues/4701), which concerns the default root-provider policy.

satanTime:
## Why

A bare module imported after its `forRoot()` result discards that result's providers when TestBed contains a mix of real and mocked imports. Repeating `.keep()` or `.mock()` for the same module also loses those providers because the builder clears its previous list and restores it only when the latest call supplies providers.

This occurs before the root provider placement handled by #4613.

## What

Retain the previous provider list across repeated calls in the same mode, appending newly supplied providers in their existing order. Add real Angular and mixed TestBed regressions for both import orders, provider identity, lazy factory execution, and multi-provider order, together with focused builder coverage.

## Impact

Repeated imports preserve configured providers and match the corresponding real Angular behavior. Deliberate keep/mock/exclude transitions and explicit provider precedence retain their existing behavior. No public API or documentation changes are needed.

Fixes #14929

### DIFF
--- a/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.spec.ts
+++ b/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.spec.ts
@@ -190,6 +190,102 @@ describe('MockBuilderPromise', () => {
     );
   });
 
+  const modes: Array<'keep' | 'mock'> = ['keep', 'mock'];
+  for (const mode of modes) {
+    for (const providerFirst of [true, false]) {
+      it(`preserves ${mode} providers with the bare module ${providerFirst ? 'last' : 'first'}`, async () => {
+        const value = { label: 'configured' };
+        const token = new InjectionToken<typeof value>('configured');
+        const providers = [{ provide: token, useValue: value }];
+        const moduleWithProviders = {
+          ngModule: TargetModule,
+          providers,
+        };
+        const builder = MockBuilder().keep(token);
+
+        if (providerFirst) {
+          builder[mode](moduleWithProviders)[mode](TargetModule);
+        } else {
+          builder[mode](TargetModule)[mode](moduleWithProviders);
+        }
+        await builder;
+
+        expect(mockHelperGet(token)).toBe(value);
+        expect(moduleWithProviders.providers).toBe(providers);
+        expect(providers).toEqual([
+          { provide: token, useValue: value },
+        ]);
+      });
+    }
+
+    it(`discards ${mode} providers when changing the module mode`, async () => {
+      const token = new InjectionToken<string>('discarded');
+      const nextMode = mode === 'keep' ? 'mock' : 'keep';
+      await MockBuilder()
+        [mode]({
+          ngModule: TargetModule,
+          providers: [{ provide: token, useValue: 'discarded' }],
+        })
+        [nextMode](TargetModule)
+        .keep(token);
+
+      expect(() => mockHelperGet(token)).toThrowError(
+        /Cannot find an instance/,
+      );
+    });
+  }
+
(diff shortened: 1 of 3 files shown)


satanTime:

Issue #3095, opened 2022-07-14
Bug: touch recursion


satanTime:
### Description of the bug

forwardRef can cause issues

satanTime:
[v14.1.0](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.1.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14962, opened 2026-09-12
fix(TestBed): resolve cyclic imports and forward references #14930

### LINKED ISSUE #14930: Check cached dependency reachability when entering a cycle from different declarations
## Why

Follow-up to #3095 and its [cycle-guard fix](https://github.com/help-me-mom/ng-mocks/commit/793a3c53700f21f31bccca7d3df3c56f0cfbd6be). The original regression checks that a self-referencing provider declaration does not recurse forever. A cycle can also leave incomplete dependency information cached for a different entry point.

In [`generateTouches`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts#L107), each declaration gets an empty `__ngMocksTouches` set before its dependencies are traversed. Callers copy the values currently in that set. For a graph `A -> [B, C]` and `B -> A`, the visit from B back to A can observe A before C has been discovered. Completing A later does not update B's cached set. A later traversal starting at B may therefore omit C and its descendants.

This is a source-supported risk requiring reproduction, not a confirmed runtime failure. Reachability matters beyond avoiding recursion: the same file uses touches to decide whether a global provider override applies and whether a module-scoped provider belongs to the configured test.

## What to check and fix

- Reproduce a supported mutual-dependency graph with an additional dependency discovered after the back-edge. One candidate follows the standalone component cycle already used by `tests/issue-6143`: A imports B and then ProviderModule; B imports A through `forwardRef`. Do not use circular NgModule imports, which Angular rejects independently.
- Have ProviderModule explicitly provide a root-scoped service through `useValue(marker)` while `ngMocks.globalMock(Service)` is active. Warm ordinary `TestBed.configureTestingModule` through A, reset, and configure through B; compare the injected marker identity with cold/reversed-order controls. Use ordinary TestBed here, because MockBuilder supplies known touches and can hide this cache-generation path.
- Assert that dependency discovery is complete and independent of traversal/cache order. Pair focused reachability assertions with an observable global-provider-policy case so the test establishes practical impact.
- If a failure is reproduced, repair cached dependency propagation or traversal without reintroducing infinite recursion. Cover self-cycles, mutual cycles, a shared acyclic dependency, and a dependency outside the configured graph.
- Preserve explicit provider precedence, global mock/keep behavior and the existing #3095 regression. Gate real Angular cases at their actual supported boundary, including the framework patch constraints already documented by #6143.

## Existing coverage

[`tests/issue-3095/test.spec.ts`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7b9e590ff6d4dd7569a3ffd5f518f2bd/tests/issue-3095/test.spec.ts) checks successful rendering without stack overflow. The newer graph traversal and cyclic mock-construction regressions cover different helpers. They do not establish complete cached touches when a later test enters a previously visited cycle from another node.

This is distinct from #14911, which tracks exception cleanup of resolver/cache stacks during failed mock construction.

satanTime:
## Why

Ordinary TestBed setup can replace an explicitly provided service with its global mock when a later test enters a dependency cycle through another declaration. The dependency cache copies incomplete transitive results during recursion and misses imports expressed through `forwardRef`, including wrapped `ModuleWithProviders` results.

This extends the cycle handling covered by #3095.

## What

Cache direct metadata dependencies and traverse them with the existing visited set. Resolve forward references through the existing library helper before inspecting module wrappers and declaration tokens. Add focused reachability regressions, real Angular cycle cases covering cold, warm, and reversed entry orders, and classic NgModule cases with forwarded `forRoot()` results.

## Impact

Explicit module providers remain reachable regardless of traversal order. Self-cycles, shared dependencies, invocation-specific `ModuleWithProviders` providers, and global mock/keep policy outside the configured graph retain their behavior. No public API or documentation changes are needed.

Fixes #14930


satanTime:

PR #14964, opened 2026-09-13
test(standalone): cover omitted declaration defaults #14932

### LINKED ISSUE #14932: Retain compiled standalone-default coverage with omitted application metadata
## Background

Follow-up to [#10306](https://github.com/help-me-mom/ng-mocks/issues/10306) and [PR #10751](https://github.com/help-me-mom/ng-mocks/pull/10751), which corrected `MockRender` after Angular 19 made declarations standalone by default. The report specifically failed after the omitted `standalone` flag changed meaning.

The dedicated [current regression](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-10306/test.spec.ts#L5-L25) explicitly sets `standalone: false`. It protects classic wrapper behavior, but no longer establishes the original application-source case where the flag is omitted and Angular's compiler supplies the new default.

Later #10762 / #10752 correctly moved standalone detection to compiled metadata. [The collector](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/resolve/collect-declarations.ts#L333-L348) reads component, directive and pipe definitions, and [isStandalone](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/common/func.is-standalone.ts#L8-L14) routes their effective value. [The ngx-file-drop regression](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests-e2e/src/issue-10762/test.spec.ts) covers older compiled declarations, while [the ngx-translate regression](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests-e2e/src/issue-10762/ngx-translate.spec.ts) uses explicitly true/false hosts and a precompiled pipe.

This follow-up is a preservation-coverage gap, not a demonstrated runtime bug or a claim that compiled standalone routing is missing. The distinction to restore is newly compiled application declarations whose decorators omit the flag.

- Add an Angular 19+ compiled fixture containing a component, an attribute directive, and a pipe with `standalone` omitted. Import them through a real host and exercise the shared `MockBuilder`/generated-declaration routing. Assert rendered input/pipe behavior and intended real/mock identities, rather than only a no-throw assertion.
- Keep a real rendering control and a mocked dependency route so both reading the original compiled default and preserving it on generated mocks are covered. Confirm the declaration is actually standalone before exercising the route, using compiler metadata assertions rather than adding `standalone: true` to make setup pass.
- Retain the existing explicit-false #10306 regression and older compiled-library controls. Explicitly nonstandalone declarations must continue to work in NgModule declarations.
- Keep generated `MockRender` wrappers explicitly nonstandalone and preserve empty/custom-template rendering. Do not rewrite every existing example or extend the compatibility claim to Angular versions before this default changed.

Use static imports and the existing `standaloneDefaultTrue` feature gate in `test-spread.conf` (Angular 19+) for the omitted-default application case. A focused source-unit check can assert that compiled `true` and `false` values survive absent annotation values across the three supported declaration kinds; real compiled execution is required for the actual default-boundary regression. Change production code only if the new test exposes a failure.

satanTime:
## Why

The standalone-default fixes in #10306, #10762, and #10752 need regression coverage for application declarations that omit `standalone`. Angular 19 and later compile those components, directives, and pipes as standalone, while older compiled declarations can still carry `false`.

## What

Add collector tests for compiled `true` and `false` across all three declaration kinds, including cache identity and preservation of the original metadata. Add Angular 19+ rendering regressions for real and mocked dependencies, input updates, projection, and a selectorless host that exercises declaration cloning.

The rendering cases also verify that generated mocks remain standalone and that empty, typed, and custom-template render wrappers remain nonstandalone.

## Impact

The existing implementation already handles these cases. This PR preserves that behavior with focused regression coverage; no public API or documentation changes are needed.

Fixes #14932

### DIFF
--- a/libs/ng-mocks/src/lib/resolve/collect-declarations.spec.ts
+++ b/libs/ng-mocks/src/lib/resolve/collect-declarations.spec.ts
@@ -1,10 +1,75 @@
 import { Component, EventEmitter, Output } from '@angular/core';
 
+import coreDefineProperty from '../common/core.define-property';
 import funcGetGlobal from '../common/func.get-global';
+import { isStandalone } from '../common/func.is-standalone';
 
 import collectDeclarations from './collect-declarations';
 
 describe('collect-declarations', () => {
+  describe('standalone', () => {
+    let reflectionOverride: PropertyDescriptor | undefined;
+
+    beforeEach(() => {
+      reflectionOverride = Object.getOwnPropertyDescriptor(
+        funcGetGlobal(),
+        '__ngMocksReflectComponentType',
+      );
+      coreDefineProperty(
+        funcGetGlobal(),
+        '__ngMocksReflectComponentType',
+        false,
+      );
+    });
+
+    afterEach(() => {
+      if (reflectionOverride) {
+        Object.defineProperty(
+          funcGetGlobal(),
+          '__ngMocksReflectComponentType',
+          reflectionOverride,
+        );
+      } else {
+        delete funcGetGlobal().__ngMocksReflectComponentType;
+      }
+    });
+
+    for (const [ngMetadataName, property] of [
+      ['Component', 'ɵcmp'],
+      ['Directive', 'ɵdir'],
+      ['Pipe', 'ɵpipe'],
+    ]) {
+      for (const standalone of [true, false]) {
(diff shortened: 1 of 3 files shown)


satanTime:

Issue #3887, opened 2022-10-22
Bug: A15 fails


satanTime:
https://app.circleci.com/pipelines/github/help-me-mom/ng-mocks/8782/workflows/65c8e896-07cd-44a8-93e4-7b899705230f/jobs/388630

satanTime:
[v14.3.1](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.3.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14965, opened 2026-09-13
test(MockModule): cover environment provider wrappers #14933

### LINKED ISSUE #14933: Cover module-wrapped environment providers through MockModule and MockBuilder
## Context

Follow-up to [#3887](https://github.com/help-me-mom/ng-mocks/issues/3887), its [Angular 15 provider-type fix](https://github.com/help-me-mom/ng-mocks/commit/2769644a545147ea57398c20ca1074a807ced5e4), and the [subsequent EnvironmentProviders compatibility change](https://github.com/help-me-mom/ng-mocks/commit/bcb81123abcac3cf9f237c77d9a097084e46c8ca).

The original change allowed `ModuleWithProviders.providers` to contain Angular's environment-provider wrappers. Current provider processing recursively unwraps them, but the retained tests do not establish module-mocking behavior for the public wrapper shape.

## Coverage gap

[`flatten`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/common/core.helpers.ts#L38) traverses both arrays and `ɵproviders` wrappers. [`mockNgDef`](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-module/mock-ng-def.ts#L13) then resolves the individual providers, so wrapping must preserve the same keep/mock decisions, order, and side-effect behavior as an ordinary provider array.

The existing [ModuleWithProviders units](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.spec.ts#L141) cover ordinary multi-provider arrays. Current `provideRouter` regressions use `.provide(...)`, which supplies explicit providers rather than exercising provider mocking inside a module or its `forRoot()` result. There is no focused `makeEnvironmentProviders` or `importProvidersFrom` fixture in the reviewed suites.

This is a coverage gap, not a claim of a current runtime failure.

## Acceptance

- Create a small real `ModuleWithProviders` fixture whose `providers` include nested `makeEnvironmentProviders(...)` wrappers and ordinary provider arrays. Consume it through direct `MockModule` and through `MockBuilder`, retaining the public Angular wrapper rather than replacing it with a hand-written private shape.
- Verify that a mocked service is a mock and its real constructor/factory has no side effects. In a kept setup, verify real provider identity and ordered multi-provider values. An explicit per-token mock in the kept setup must still win.
- Include one environment initializer to establish that kept initialization runs once against the final provider configuration, while the mocked setup does not accidentally execute an omitted real initializer. Keep this at the environment/module injector level, where Angular permits these providers.
- Compare with an equivalent ordinary provider-array control and confirm that processing does not mutate the source wrapper or its arrays. Retain a direct `.provide(...)` control only if needed to distinguish explicit-provider passthrough from module mocking.
- Cover the recursive unwrapping branch with focused source assertions and establish the public types and runtime behavior in compiled Angular targets at the APIs' actual introduction boundaries.

Investigate and fix a shared provider defect only if these regressions expose one. This does not request root-provider policy changes or alternate recipes in `@Injectable` metadata, which are separate from module-contained environment wrappers.

satanTime:
## Why

The environment-provider compatibility changes following #3887 need coverage for provider mocking inside modules. Explicit providers passed to a test do not exercise the same keep/mock decisions as nested wrappers in `ModuleWithProviders`.

## What

Compare ordinary provider arrays with nested `makeEnvironmentProviders` and `importProvidersFrom` wrappers through direct `MockModule`, mocked `MockBuilder`, and kept modules with an explicit token mock. Cover real/mock identity, ordered multi-providers, suppressed constructor and factory effects, and exactly-once kept initialization against the final providers.

Add an Angular 14+ `importProvidersFrom` environment-injector control and focused flattening tests for order, leaf identity, repeated wrappers, and source preservation. Gate module-contained environment wrappers at Angular 15.

## Impact

The existing implementation handles these cases. This PR adds regression coverage without changing runtime behavior, public APIs, or documentation.

Fixes #14933


satanTime:

Issue #6356, opened 2023-07-24
Docs: Rename testing guide NGRS to NGXS


mfrieling:
There's a repeated typo in the docs (testing guides): The second state management library besides NGRX is **not NGRS** but NGXS**. Due to this typo I couldn't find it and had to find out how to test the state on my own with a lots of problems and a way more complex final result than in the docs.

In addition, the example doesn't work for me, I get the error message
> Cannot find an instance via ngMocks.findInstance(Store)

I'm using 
- Angular: 16.1.6 **(Standalone)**
- @nx/angular: 16.3.2 (NX workspace with Angular preset and module federation)
- @ngxs/store: 3.8.1
- Jest: 29.5.0
- ng-mocks: 14.11.0

satanTime:
Hi @mfrieling,

thanks for noting, such a silly typo...

satanTime:
[v14.12.0](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.12.0) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14967, opened 2026-09-13
test(NGXS): cover and correct feature-only setup #14935

### LINKED ISSUE #14935: Cover the documented NGXS feature-only module setup with a standalone consumer
## Background

Follow-up to [#6356](https://github.com/help-me-mom/ng-mocks/issues/6356) and its [documentation correction](https://github.com/help-me-mom/ng-mocks/commit/8e23d445d). In addition to the NGXS name typo, the original report described a `Cannot find an instance via ngMocks.findInstance(Store)` error in an Angular 16 standalone application.

The current [NGXS guide](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/docs/articles/guides/libraries/ngxs.md#L24-L38) documents a separate setup for a feature module that imports only `NgxsModule.forFeature`: add `NgxsModule.forRoot()` and `Store` through the kept inputs and keep the feature module type.

The [existing integration](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests-e2e/src/ngxs/test.spec.ts#L52-L109) instead declares a classic component in a module that already imports `forRoot([TestState])` and an empty `forFeature()`. It verifies root-state dispatch/select behavior, but does not execute the documented feature-only recipe or register its tested state through `forFeature`.

This is missing coverage of an independent documented setup, not a reproduced new NGXS or ng-mocks failure.

## Bounded regression and documentation acceptance

- Add a focused NGXS integration with the state registered only through `NgxsModule.forFeature([TestState])`. Its application feature module must not call `forRoot`.
- Supply `NgxsModule.forRoot()` and the real `Store` through `MockBuilder` as shown in the guide. Keep the feature registration, and use a standalone consumer relevant to the original report.
- Assert `ngMocks.findInstance(Store)` / TestBed lookup resolves the intended real store, dispatch reaches the real feature action handler, and selecting the resulting state updates the rendered consumer. Verify the actual feature-state value; a successful component creation alone would not establish this setup.
- Keep `Store.dispatch` real for this case; copying the existing test's dispatch stub would bypass the feature action being checked. Render before looking up the store, or preconfigure `MockRenderFactory`, so an early TestBed injection does not trigger an unrelated MockRender flush error.
- Preserve the existing classic root-module integration as a control. Keep the investigation to the documented NgModule-with-providers recipe; do not expand it into unrelated plugin, lazy-loader, or modern provider-API combinations.
- Compare the guide's feature-only snippet with the executable fixture. If that recipe needs correction, update it with the smallest working setup and explain its root-versus-feature provider requirement. Otherwise retain the public behavior and add the missing regression.

Change production behavior only if the focused integration exposes a defect. Use the existing integration runner and the installed NGXS/Angular compatibility boundary.

satanTime:
## Why

The feature-only NGXS recipe places `Store` among MockBuilder's test targets, which mocks its internal root dependencies. Its separate root import also follows the feature import, causing an invalid initialization order. The existing classic integration does not exercise this setup or real feature dispatch, leaving the standalone consumer case raised in #6356 uncovered.

## What

Correct the guide to use a test module that imports NGXS root services before the application feature module, with `Store` and the NGXS module types kept as dependencies. Add real TestBed and MockBuilder standalone-consumer regressions that check Store lookup, real feature dispatch, selected state, rendered updates, and the absence of initialization errors.

## Impact

Feature-only applications have an executable setup that retains real NGXS behavior while application dependencies can still be mocked. The existing classic integration remains unchanged. No production code or public API change is needed.

Closes #14935

### DIFF
--- a/docs/articles/guides/libraries/ngxs.md
+++ b/docs/articles/guides/libraries/ngxs.md
@@ -21,19 +21,20 @@ beforeEach(() =>
  );
 ```
 
-If your module imports only `NgxsModule.forFeature`, you need to add `NgxsModule.forRoot()` and `Store` manually:
+If your module imports only `NgxsModule.forFeature`, use a test module to initialize
+`NgxsModule.forRoot()` before the feature module. Keep `Store` as a dependency so its
+NGXS root services stay real:
 
 ```ts
+@NgModule({
+  imports: [NgxsModule.forRoot(), TargetModule],
+})
+class TestModule {}
+
 beforeEach(() =>
-   MockBuilder(
-     // keep and export
-     [
-       TargetComponent,
-       NgxsModule.forRoot(), // provides required services
-       Store, // keeps the root provider of the store
-     ],
-     // mock
-     TargetModule,
-   ).keep(NgxsModule.forFeature().ngModule) // keeps all NgxsModule.forFeature
+   MockBuilder(TargetComponent, TestModule)
+     .keep(Store)
+     .keep(NgxsModule.forRoot().ngModule)
+     .keep(NgxsModule.forFeature().ngModule)
  );
 ```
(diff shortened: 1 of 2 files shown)


satanTime:

Issue #4641, opened 2023-01-09
Bug: NG0300 thrown when MockModule and TestBed.overrideModule are used together


ma7moudat:
### Description of the bug

This occurred while trying to update from v14.4.0 to v14.5.2.

#### Scenario:

- `IconComponent` in `IconModule`.
- `Dummy1Component` in `Dummy1Module`. Imports `IconModule` and renders `IconComponent`.
- `Dummy2Component` in `Dummy2Module`. Imports `IconModule` and renders `IconComponent`.
- `MainComponent` the component to be tested. Renders `Dummy1Component`, `Dummy2Component` and `IconComponent`.
- In the test:
    - `IconComponent` is replaced by `IconMockComponent` using `TestBed.overrideModule()`.
    - `Dummy1Module` is mocked with ng-mocks' `MockModule()`.

#### Result:
An error is thrown:

```sh
Error: NG0300: Multiple components match node with tagname app-icon: IconMockComponent and IconComponent. 
Find more at https://angular.io/errors/NG0300
...
```

Once the `MockModule()` call or the `TestBed.overrideModule()` call is removed, the test passes.

### An example of the bug

Link: https://codesandbox.io/s/gallant-ully-kof3s7?file=/src/test.spec.ts

### Expected vs actual behavior

Not sure TBH, since this error is thrown by Angular by design, but is there a way for the mock from `TestBed.overrideModule()` to take precedence?

In our project we use `TestBed.overrideModule()` in the test setup script to apply some global component mocks and in the tests, we use `MockModule()` for convenience.

satanTime:
[v14.5.3](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.5.3) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

PR #14970, opened 2026-09-13
fix(TestBed): preserve mixed import order #14937

### LINKED ISSUE #14937: Check nested kept consumers against the Angular module-override baseline
Follow-up to [#4641](https://github.com/help-me-mom/ng-mocks/issues/4641) and its [mixed TestBed export fix](https://github.com/help-me-mom/ng-mocks/commit/66a8ecc40).

## Unresolved preservation coverage

The existing regression overrides a module's exported component while TestBed mixes real modules with an explicit `MockModule`. It proves that a root consumer sees the replacement and that a mocked dependency has an empty template. However, the [assertion for a kept nested consumer](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-4641/test.spec.ts#L100) remains commented out with a tentative explanation that Angular does not redefine nested imports.

That leaves an important part of the original preservation contract unestablished: whether a real consumer reached through a real module sees the same supported `TestBed.overrideModule` result when an unrelated mocked module triggers ng-mocks' [mixed TestBed reconstruction](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts#L266).

This is a bounded baseline/coverage investigation, not a claim of a current ng-mocks defect. The old comment may describe a framework limitation; the test must determine that before assigning blame or requiring a production fix.

- Start with real TestBed modules and `TestBed.createComponent`, using the retained topology: `ReplaceModule` exports an original component, and a real nested consumer imports `ReplaceModule`. Override the declarations and exports with the replacement component. Establish which supported Angular profiles propagate that override into the nested consumer.
- Add the unrelated explicit `MockModule` boundary that triggers mixed reconstruction, keeping the nested consumer real. On profiles where the baseline supports the override, assert both root and nested consumers render the replacement and expose the replacement instance identity.
- Assert the original replaced component's constructor does not run, while the unrelated mocked consumer remains mocked and its original constructor/template behavior stays suppressed. Use distinct rendered markers so a root-only replacement cannot satisfy the nested assertion.
- Reverse the real/mock top-level import order as a preservation control. Retain the current [root replacement case](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-4641/mixed.spec.ts#L92).
- Replace the tentative commented-out expectation with an executable assertion or a precise, supported-profile limitation based on the real Angular baseline. Fix shared reconstruction only if it diverges from that baseline; do not require ng-mocks to implement an override Angular itself does not support.

The later [mixed-boundary regressions](https://github.com/help-me-mom/ng-mocks/blob/c598f25b7/tests/issue-14895/test.spec.ts#L244) cover shared real descendants and import order, but do not call `overrideModule`. This is also separate from [#14914](https://github.com/help-me-mom/ng-mocks/issues/14914), which concerns isolation of inherited Ivy definitions in generated declarations.

satanTime:
## Why

Mixed `TestBed.configureTestingModule` setups rebuild imports through MockBuilder, which places mocked modules before kept modules. Ivy's shared-module override traversal depends on import order: moving a mocked module ahead of a real nested consumer can leave that consumer using the original declaration after `TestBed.overrideModule`.

This establishes the missing nested-consumer baseline from #4641 and fixes the divergence in #14937.

## What

- Preserve explicit import order when passing reconstructed metadata to TestBed, retaining generated module identities, provider wrappers, and internal imports. Repeated module configurations retain their existing merged-provider precedence.
- Add independent real and mixed Angular regressions for root and nested replacement rendering and identity, with original-constructor and mocked-boundary suppression checks.
- Cover reversed ordering, nested import arrays, empty entries, standalone imports, and module-provider preservation in focused source tests. An Angular injection regression also protects explicit providers when a global replacement adds another import of the same module. Enable the original #4641 nested assertion using the order supported by Angular's baseline.

## Impact

Existing mixed TestBed setups retain the nested override behavior supported by their real-module equivalents. This restores the existing contract, so public documentation does not need to change.

Fixes #14937

### DIFF
--- a/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.imports.spec.ts
+++ b/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.imports.spec.ts
@@ -1,0 +1,486 @@
+import {
+  Component,
+  Injectable,
+  InjectionToken,
+  NgModule,
+} from '@angular/core';
+import { TestBed, TestModuleMetadata } from '@angular/core/testing';
+
+import funcExtractTokens from '../mock-builder/func.extract-tokens';
+import { MockBuilderStash } from '../mock-builder/mock-builder-stash';
+import { EntryComponentsModule } from '../mock-builder/promise/handle-entry-components';
+import { MockComponent } from '../mock-component/mock-component';
+import { ngMocks } from '../mock-helper/mock-helper';
+import mockHelperFasterInstall from '../mock-helper/mock-helper.faster-install';
+import { MockModule } from '../mock-module/mock-module';
+
+import { flatten } from './core.helpers';
+import { getSourceOfMock } from './func.get-source-of-mock';
+import { isMockedNgDefOf } from './func.is-mocked-ng-def-of';
+import { isNgDef } from './func.is-ng-def';
+import { isNgModuleDefWithProviders } from './func.is-ng-module-def-with-providers';
+
+import './ng-mocks-global-overrides';
+
+// @see https://github.com/help-me-mom/ng-mocks/issues/14937
+describe('ng-mocks-global-overrides:imports', () => {
+  const stash = new MockBuilderStash();
+
+  beforeEach(() => {
+    ngMocks.flushTestBed();
+    stash.backup();
+    expect(mockHelperFasterInstall().before.length).toBe(1);
+  });
+
+  afterEach(() => {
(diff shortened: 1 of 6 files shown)


satanTime:

PR #14983, opened 2026-09-13
fix(MockRender): apply overrides to render middleware #14945

### LINKED ISSUE #14945: Check late TestBed strategy overrides in View Engine middleware clones
While investigating #14915, source review suggested a related gap in Angular 5–11 View Engine. This has not yet been reproduced at runtime.

`MockRenderFactory` immediately creates a separately decorated middleware class for a selectorless component. A subsequent `TestBed.overrideComponent(original, ...)` may update only the original class, leaving the clone's captured change-detection strategy unchanged. The compiled-definition refresh used by Ivy does not apply to View Engine.

Check both explicit `Default` → `OnPush` and `OnPush` → `Default` transitions applied after factory creation but before rendering. Use a classic NgModule component with an array input and render its length. After mutating the same array reference, expect unchanged text for OnPush and updated text for Default; replacing the array should update both.

Compare overrides applied before factory creation, a normal simple-selector component, and direct TestBed rendering. If a failure is confirmed, check whether the same original-versus-clone lookup also affects template, provider/view-provider, and host overrides. Preserve original metadata, provider aliases, input identity, and subsequent use of the original component.

Relevant paths: `mock-render-factory.ts`, `func.reflect-template.ts`, and `func.inherit-definition.ts` in `libs/ng-mocks/src/lib/mock-render`. Related #1876 covers selectorless providers but does not establish this override timing and View Engine behavior.

satanTime:
## Why

Selectorless components and directives with complex selectors use separately decorated render middleware. Direct TestBed overrides target the original declaration, leaving the middleware with copied metadata. This loses early and late change-detection overrides in View Engine and also affects component templates, providers, view providers, and directive providers.

## What

- Forward accepted component and directive overrides to active middleware through Angular's native metadata override handling, preserving operation order, generated selectors, and injection aliases.
- Restore the existing inheritance handling after Ivy recompiles middleware, keeping host listeners and bindings from being applied twice. Clear override tracking with the normal TestBed reset.
- Cover both strategy directions before and after factory creation, simple-selector controls, component and directive metadata, ordered provider additions/removals, provider clearing, and reuse of the original declarations after reset.

## Impact

Existing TestBed overrides affect MockRender middleware across the supported rendering engines while original annotations, input references, provider identities, and listener behavior remain intact. The fix restores expected behavior, so public documentation is unchanged.

Fixes #14945

### DIFF
--- a/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts
+++ b/libs/ng-mocks/src/lib/common/ng-mocks-global-overrides.ts
@@ -20,6 +20,7 @@ import { resetDeclarationFactories } from './ng-mocks-declaration-factories';
 import { rememberMockDeclarations, resetInjectedDeclarations } from './ng-mocks-injected-declarations';
 import { applyPlatformOverrides, defineTouches } from './ng-mocks-platform-overrides';
 import { resetRuntimeInject } from './ng-mocks-runtime-inject';
+import { installTemplateOverrides, resetTemplateOverrides } from './ng-mocks-template-overrides';
 import { installTestBedInjection } from './ng-mocks-test-bed-injection';
 import { rememberTestModuleOptions, resetTestModuleOptions } from './ng-mocks-test-module-metadata';
 import ngMocksUniverse from './ng-mocks-universe';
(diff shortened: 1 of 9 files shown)


satanTime:

PR #14989, opened 2026-09-13
fix(ngMocks): resolve remapped host inputs in older Ivy metadata #14956

### LINKED ISSUE #14956: Check remapped host input lookup with older Ivy metadata
## Context

Related to #14923 and #14955. The shared-alias ownership audit also found a separate possible gap when a host directive exposes an input under a different alias on older Ivy metadata layouts.

## Suspected defect

Angular 15–17 host-directive mappings are stored in the ordinary node input map under the exposed alias. The entries retain the host directive's private property name. Reflection on the host directive itself retains its original public input alias, without the containing declaration's remapping.

The value crawler selects the exposed node alias but resolves reflected inputs by that alias. For a host input declared as `@Input('original') backingField` and exposed through `inputs: ['original: exposed']`, neither the reflected alias nor the direct `instance.exposed` fallback identifies `backingField`.

This is a source-supported suspicion, not a captured runtime failure. Newer Angular versions with `hostDirectiveInputs` supply the original public alias separately; they should be controls in the investigation.

- Reproduce with real, kept host directives and remapped aliases on Angular 15 and 17. Determine the exact metadata-version boundary and include the latest Angular target as a control.
- Exercise `reveal`, `revealAll`, `findTemplateRef`, and `findTemplateRefs` on an actual `ng-template`; check stored values, updates, identity, duplicate-free results, and missing values including `undefined`.
- Include a decoy instance property named after the exposed alias so a raw-property fallback cannot hide the defect.
- Check more than one host/ordinary directive exposing the same outer alias, and preserve unremapped aliases, transformed values, and signal input behavior where supported.
- Add focused source-unit coverage for the affected tuple layouts and a real-Angular regression before changing source. Keep the fix within existing lookup paths and avoid new public APIs or options.

satanTime:
## Why

When a host directive exposes an aliased input under another name, Angular 15 through 19.1 stores the private backing name in the node's input tuples. Attribute/value lookup used the exposed alias instead, so it could miss the actual input or match an unrelated property with that name. This is a follow-up to the shared input lookup fix in #14955.

## What

Resolve older node tuples by their private input names while preserving public-alias resolution for newer ordinary and host directive metadata. Add focused metadata tests and real template regressions covering remapped decorator and signal inputs, shared aliases, live values, and signal transforms.

## Impact

`ngMocks.reveal`, `ngMocks.revealAll`, `ngMocks.findTemplateRef`, and `ngMocks.findTemplateRefs` find the stored input value through a remapped host alias. Lookup preserves node and template identity, returns each match once, and does not rerun input transforms. The existing public contract is restored, so API and documentation changes are unnecessary.

Fixes #14956


satanTime:

Issue #14925, opened 2026-09-11
Feature Request: a how-to for testing Angular 22 signal forms


kekel87:
### Describe the feature or problem you'd like to solve

Hi, it's me again 🙃 !

Signal forms are stable in Angular 22 and we're preparing our app's migration. There's nothing about them on the site yet: the `ngMocks.change` page and the mock form controls guide are entirely `ControlValueAccessor` based, and the package has no signal forms symbol (no `FieldTree`, no `FormValueControl`, no `FORM_FIELD`). So we found everything by trial and error, and it might be worth a page.

The three things that cost us the most time:

**1. The spec needs `.keep(NG_MOCKS_ROOT_PROVIDERS)`.** Without it, dispatching an `input` event on a control bound with `[formField]` fails with `TypeError: validityMonitor.isBadInput is not a function`. Angular 22 injects a root `InputValidityMonitor` and `MockBuilder` mocks it by default. The tricky part is that the error comes out of the listener and not out of the call, so all the spec sees is an empty model 😵

**2. A `FieldTree` doesn't go through a `MockRender` template string.** The component form works, and has since 14.16.0 (#13089). The template string form still fails on HEAD, in three different ways depending on what we pass:

```ts
MockRender('<input [formField]="f.name" />', { f });             // result.replace is not a function
MockRender('<input [formField]="f.firstname" />', { f });        // this.field(...) is not a function
MockRender('<input [formField]="field" />', { field: f.name });  // Reflect.getOwnPropertyDescriptor called on non-object
```

A `FieldTree` is a callable proxy, so `funcGetName` (`common/func.get-name.ts:18`) reading `value.name` gets a child `FieldTree` instead of a string when a field is actually named `name`. We have about fifty spec files on the template string form, so a simple "use the component form" would have saved us time. Do you want this one as a separate ticket? I didn't want to open three issues at once 🙂

**3. Selectors change.** `formControlName` disappears from the DOM, and the `name` attribute only exists when `FormField` is kept in the `MockBuilder` list, so `input[name="contact.firstname"]` silently matches nothing when it's mocked. On a component host, `[formField]` sets no `name` attribute at all.

### Proposed solution

A "Testing signal forms" page, plus a line on the `ngMocks.change` page saying which hosts are supported. Even just this table would cover most cases:

| `[formField]` host | what works today |
| --- | --- |
[...]

PR #14991, opened 2026-09-13
fix(ngMocks.change): handle native form control values #14985

### LINKED ISSUE #14985: docs(signal-forms): explain field selectors when migrating tests
Parent request: #14925.

## Why

Tests migrated from classic forms can silently find no elements when they keep selectors based on `formControlName` or assume that every `[formField]` host receives a DOM `name` attribute. The current signal-forms guide uses a single native `input` or a component type, but does not explain these selector changes or how to distinguish multiple fields.

## What

- Add concise selector guidance to `docs/articles/guides/signal-forms.md` with a before/after migration example.
- Explain that a signal-form binding does not preserve the former `formControlName` attribute.
- Verify and document that generated native `name` attributes depend on the real `FormField` directive; mocking it does not preserve its host behavior.
- Explain that binding a custom component's optional `name` input does not automatically add a DOM `name` attribute to its host.
- Show explicit `data-testid` selectors for multiple native fields and component-type or explicit-attribute selectors for custom controls. Avoid relying on generated field-name formatting.

## Completion criteria

- [ ] The guide explains the failed selector patterns reported in #14925 and provides concrete replacements.
- [ ] Executable examples under `examples/TestSignalForms` substantiate the native and custom-host behavior and show that an interaction updates the intended field while leaving a sibling unchanged.
- [ ] The guidance distinguishes selecting a field from keeping the real form binding needed to exercise form behavior.

## References

- [Existing guide and examples](https://github.com/help-me-mom/ng-mocks/blob/main/docs/articles/guides/signal-forms.md)
- [Angular FormField native-property handling](https://github.com/angular/angular/blob/main/packages/forms/signals/src/directive/form_field.ts)
- [Angular custom-control input binding](https://github.com/angular/angular/blob/main/packages/forms/signals/src/directive/control_custom.ts)

satanTime:
## Why

The form guides mixed testing a real form with mocking a child control, and did not show a consistent way to find a host, read its value, change it, and assert the result. Plain native inputs could throw a missing-accessor error, multiple selects did not apply arrays correctly, and radio changes could leave the host's checked state inconsistent with the selected value.

## What

Correct `ngMocks.change` for plain native inputs, textareas, selects, and multiple selections. Use Angular's built-in multiple-select accessor for model values, including object options and `compareWith`, without adding `writeValue` calls to custom or mocked accessors. Boolean arguments set the checked state of native radios and checkboxes, regardless of the option's value. Preserve the legacy handling of nonboolean arguments. Boolean changes keep option values intact and avoid emitting a radio selection when unchecking a host. Inspect actual input bindings without constructing unrelated providers.

Add focused guides under **How to test** for native inputs, ngModel, reactive forms, signal forms, and submission. Keep the component and template first, separate the MockBuilder setup, and teach find → read → change → assert with executable input-type examples.

Show direct CSS or component selectors in simple `ngMocks.change` calls, with an element-variable alternative when the test already inspects that host. Correct the `change` and `touch` parameter types to accept the component selectors their lookup already supports. Reuse `ngMocks.reveal` results for Angular binding lookups. Keep concise explanations of setup and timing in the articles and executable examples, and document that convention in the repository guidance.

Keep **How to mock** separate: cover ControlValueAccessor children, custom signal controls, and mocked form bindings. Show both value directions and explain which bindings preserve form behavior. Add executable examples and source, CodeSandbox, and StackBlitz links. Retain the existing validator API references.

## Impact

Tests can use the same helper flow across native controls and Angular form hosts. Custom-element CVAs and `FormValueControl` / `FormCheckboxControl` retain their value semantics, including boolean checked values on custom checkbox controls.

For both native checkboxes and radios, `ngMocks.change(control, true)` checks the selected host and `false` unchecks it. A classic radio bound with `[value]="false"` is checked by `true` and writes its option value `false` to the model. The guides and executable examples use this checked-state flow, with regressions preserving older nonboolean calls.

Closes #14985.
Related to #14925.

### DIFF
--- a/docs/articles/guides/signal-forms.md
+++ b/docs/articles/guides/signal-forms.md
@@ -15,6 +15,8 @@ custom controls that implement `ControlValueAccessor` or `FormValueControl`.
 
 - [`MockBuilder`](/api/MockBuilder.md)
 - [`MockRender`](/api/MockRender.md)
+- [`ngMocks.find`](/api/ngMocks/find.md)
+- [`ngMocks.findAll`](/api/ngMocks/findAll.md)
 - [`ngMocks.change`](/api/ngMocks/change.md)
 - [`ngMocks.touch`](/api/ngMocks/touch.md)
 - [`MockInstance`](/api/MockInstance.md)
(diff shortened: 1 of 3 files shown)


satanTime:

PR #14992, opened 2026-09-13
fix(MockBuilder): retain JIT root provider recipes #14949

### LINKED ISSUE #14949: Check kept root provider recipes when JIT annotations omit metadata

satanTime:
## Why

Keeping a root provider whose JIT annotations omit its provider options registers the token as a class. Angular then executes the token's constructor instead of its configured factory, implementation, alias, or value. The metadata reflection path added for #14917 cannot recover those missing options.

The related inheritance review also found that wrapping an inherited constructor factory without its child type can instantiate the parent when the child is the builder's target.

## What

Use Angular's own root injectable factory for otherwise unresolved kept class providers after module providers have been selected. Preserve the existing constructor-factory protocol by forwarding the kept child type when wrapping inherited construction.

Add JIT regressions for all four provider recipes, explicit and hidden dependencies, empty values, lazy creation, singleton and alias identity, module and explicit-provider precedence, and inherited construction. Source units cover modern and legacy native definitions, ownership, scope, and preservation of selected provider objects.

## Impact

Kept JIT root providers resolve their configured recipes while module overrides, explicit providers, and dependency-mocking decisions retain their precedence. The existing metadata reflection path remains supported. This restores the expected behavior without changing public APIs or documentation.

Fixes #14949


satanTime:

PR #15009, opened 2026-09-13
fix(MockInstance): handle every supported member key #14997

### LINKED ISSUE #14997: MockInstance misclassifies empty-string, numeric, and symbol member keys
## Problem

`MockInstance` accepts a member name through `K extends keyof T`, but some valid property keys are treated as initialization callbacks. Creating the mocked provider then throws instead of applying the customization.

This was found while reviewing the fixes for #2087 and #1256. Those fixes correctly distinguish falsy customization values from resets and unify scope handling, but the member/callback distinction still depends on a string-only parser and a truthiness check.

## Reproducer

```ts
import { MockInstance, MockProvider } from 'ng-mocks';

@Injectable()
class TargetService {
  public '' = 0;
}

describe('MockInstance property keys', () => {
  MockInstance.scope();

  beforeEach(() => TestBed.configureTestingModule({
    providers: [MockProvider(TargetService)],
  }));

  it('customizes an empty-string property', () => {
    MockInstance(TargetService, '', 42);
    expect(TestBed.inject(TargetService)['']).toBe(42);
  });
});
```

Expected: the injected mock has the requested member value. Current source routes `42` into the callback list and calls it as a function. Numeric and symbol member keys accepted by the same public overload are also misclassified, before the application step.

## Cause and scope

- [`parseMockInstanceArgs`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-instance/mock-instance.ts) recognizes a member only when its key is a string.
- [`mock-instance-apply`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-instance/mock-instance-apply.ts) uses `if (name)`, so an empty string is interpreted as a callback customization.
- [`helper.use-factory`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/helper.use-factory.ts) invokes every resulting callback.

This is source-confirmed; the proposed regression above has not yet been executed. The open live-instance feature PR #14818 leaves these parser/application paths unchanged.

## Expected fix

Correct dispatch within the existing overloads. Cover empty-string, numeric (including zero), and symbol keys for values and accessors, while preserving ordinary string keys, callback initializers, falsy values, and scope restoration. No new public parameter or mode is needed.

satanTime:
## Why

`MockInstance` accepts any member key through its existing overloads, but numeric and symbol keys were parsed as initializers. An empty string was parsed as a member and then treated as an initializer when applying the customization, causing failures or invoking a method stub during mock creation.

## What

Recognize string, numeric, and symbol member keys and distinguish an absent name from a falsy key. Add focused source and Angular regressions for values, accessors, callable stubs, symbol identity, component customization, and scope/reset behavior.

## Impact

Mock customizations now work for the member keys already supported by the public signatures. Existing callbacks, falsy values, and accessor pairs retain their behavior. Public signatures and documentation remain unchanged.

Fixes #14997


satanTime:

PR #15010, opened 2026-09-13
fix(MockRender): preserve wrapper cache section boundaries #14998

### LINKED ISSUE #14998: MockRender cache confuses providers with viewProviders
## Problem

`MockRenderFactory` can reuse a wrapper with the wrong provider scope. Moving the same provider between `providers` and `viewProviders` produces an identical cache key when the other list is explicitly empty. Rendering then depends on which factory was created first.

Related history: #731 / #732 introduced wrapper caching; #726 / #729 and #1507 added or corrected the distinct provider scopes. Recent factory isolation work in #14913 does not change this cache key.

## Reproducer

```ts
import { Component, Inject, InjectionToken, Optional, Self } from '@angular/core';
import { MockRenderFactory, ngMocks } from 'ng-mocks';

const TOKEN = new InjectionToken<object>('scope');
const value = {};
const provider = { provide: TOKEN, useValue: value };

@Component({
  selector: 'cache-scope-target',
  standalone: false,
  template: '',
})
class TargetComponent {
  constructor(@Inject(TOKEN) @Optional() @Self() public local: object | null) {}
}

it('keeps provider scopes distinct in the wrapper cache', () => {
  ngMocks.reset();
  TestBed.configureTestingModule({ declarations: [TargetComponent] });
  const providersFactory = MockRenderFactory(TargetComponent, [], {
    configureTestBed: false,
    providers: [provider],
    viewProviders: [],
  });
  const viewProvidersFactory = MockRenderFactory(TargetComponent, [], {
    configureTestBed: false,
    providers: [],
    viewProviders: [provider],
  });

  expect(viewProvidersFactory.declaration).not.toBe(providersFactory.declaration);
  expect(viewProvidersFactory().point.componentInstance.local).toBeNull();
});
```

The first factory need not render or configure TestBed. Its creation alone warms the cache. The second factory currently receives that same declaration; the generated target provider directive exposes `value` to the `Self` lookup, despite the second factory requesting only wrapper `viewProviders`.

## Cause and scope

[`func.create-wrapper.ts`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-render/func.create-wrapper.ts) concatenates template, bindings, providers, and viewProviders into a flat array, without retaining the boundaries between sections. The two configurations above both become `[TargetComponent, provider]`. Cache matching compares only total length and element identity, then returns before processing the new flags.

This is source-confirmed at audit commit `4170bcc43`; the proposed Angular regression has not yet been executed.

## Expected fix

Keep cache sections distinguishable without changing public options or disabling caching. Add both warming orders, distinguish omitted and empty lists where behavior differs, and preserve reuse for identical configurations. Verify the actual injection scope as well as wrapper identity, for both factory and ordinary MockRender callers.

satanTime:
## Why

`MockRender` and `MockRenderFactory` flattened bindings, providers, and view providers into one wrapper-cache key. Moving the same provider between scopes could reuse a wrapper with the wrong injection behavior. Empty and omitted sections could also collide, causing explicit empty bindings to inherit default input bindings, or the reverse.

## What

Retain each section's length in the cache key. Add regressions for both warming orders, factory and ordinary render calls, input defaults, and nonempty provider partitions.

## Impact

Cached renders retain their requested injection scope and input bindings. Identical configurations still reuse wrappers, with provider identity and ordering preserved. Existing public options and documentation remain unchanged.

Fixes #14998


satanTime:

PR #15011, opened 2026-09-13
fix(MockService): include symbol members in generated mocks #14999

### LINKED ISSUE #14999: Mocked symbol members still execute real methods and accessors
## Problem

`MockService` and Angular declaration mocks leave symbol-named prototype methods and accessors connected to the real implementation. Calling such a method on a mock can execute application logic or throw, although the existing MockService contract promises dummy methods.

This is a wider case of the member/accessor discovery fixes discussed in #177 and #196, and of the customization behavior documented through #270. The shared declaration callers were also checked while reviewing the inheritance fix #69 / #94. It is separate from #14916, which corrected own string accessors.

## Reproducer

```ts
import { MockService } from 'ng-mocks';

const request = Symbol('request');

class TargetService {
  public [request](): void {
    throw new Error('real method executed');
  }
}

it('replaces symbol methods with dummies', () => {
  const mock = MockService(TargetService);
  expect(() => mock[request]()).not.toThrow();
});
```

Expected: the symbol method behaves like an ordinary mocked method and does not execute the real body. Current source leaves the real method inherited from `TargetService.prototype`. A symbol getter/setter has the same problem, including real accessor side effects.

## Cause and scope

At audit commit `4170bcc430d95e001c92309cb12f7c7dfb043dc1`, [`helper.extract-methods-from-prototype`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/helper.extract-methods-from-prototype.ts) and [`helper.extract-properties-from-prototype`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/helper.extract-properties-from-prototype.ts) enumerate only `Object.getOwnPropertyNames`. [`helper.create-mock-from-prototype`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/helper.create-mock-from-prototype.ts) then attaches the original prototype to the mock, making the skipped symbol members real.

The shared [`Mock` constructor](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/common/mock.ts#L251-L257) uses the same extractors before attaching the original prototype. Symbol members therefore also escape `MockComponent`, `MockDirective`, and `MockPipe`; the pipe initializer replaces `transform` but does not replace additional symbol members.

This is source-confirmed; the proposed regression has not yet been executed.

## Expected fix

Mock discoverable symbol members through the existing APIs. Cover inherited symbol methods and accessors, the class and existing-instance MockService forms, the MockProvider path, and real Angular component/directive/pipe mock instances. Preserve symbol identity for member access, safe spy-name generation, ordinary string members, and constructor-free mocking. Avoid invoking real getters or adding a public option to enable the correction.

satanTime:
## Why

Mock discovery enumerates only string property names. Symbol methods and accessors on original prototypes can therefore execute real code through a mock, while symbols on supplied instances and overrides are omitted. The same discovery code serves service, component, directive, and pipe mocks.

## What

Preserve symbol keys during member discovery, dummy creation, override application, and provider implementation copying. Convert keys to text only for spy labels, and recognize symbol-only classes compiled to ES5. Regression tests cover each affected mock type, inherited and instance members, accessor state, overrides, and suppressed real side effects.

Preserve replacement-function metadata precedence, existing string property forwarding, and readable form-control diagnostics in the other consumers of the shared discovery helpers.

## Impact

Existing mocking APIs handle symbol members consistently with ordinary members. Constructors remain suppressed, original descriptors remain intact, and Angular inputs, outputs, and string property forwarding retain their existing behavior. This restores the documented mocking contract and requires no public API or documentation changes.

Fixes #14999


satanTime:

PR #15012, opened 2026-09-13
fix(ngMocks.guts): retain repeated module wrapper providers #15000

### LINKED ISSUE #15000: ngMocks.guts drops providers from repeated ModuleWithProviders entries
## Problem

`ngMocks.guts` discards providers from later `ModuleWithProviders` entries when they refer to the same module. This can remove a dependency entirely or ignore a later provider override.

This was found while reviewing #241, #197, and #151. The newer #14929 repair covers MockBuilder and mixed TestBed normalization; `guts` still uses a separate module-only deduplication path.

## Reproducer

```ts
import { InjectionToken, NgModule } from '@angular/core';
import { ngMocks } from 'ng-mocks';

const FIRST = new InjectionToken<string>('first');
const SECOND = new InjectionToken<string>('second');

@NgModule({})
class SharedModule {}

it('keeps providers from each wrapper', () => {
  const first = {
    ngModule: SharedModule,
    providers: [{ provide: FIRST, useValue: 'first' }],
  };
  const second = {
    ngModule: SharedModule,
    providers: [{ provide: SECOND, useValue: 'second' }],
  };

  TestBed.configureTestingModule(ngMocks.guts(SharedModule, [first, second]));

  expect(TestBed.inject(FIRST)).toBe('first');
  expect(TestBed.inject(SECOND)).toBe('second');
});
```

Expected: both tokens are available, as they are with native `imports: [first, second]`. Current source drops `second` and injection of `SECOND` fails. Keeping `SharedModule` makes both wrapper provider recipes real, so this does not depend on how automatic mocking transforms values.

## Cause and scope

[`handleModuleWithProviders` in mock-helper.guts.ts](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-helper/mock-helper.guts.ts) returns immediately when `data.skip` already contains `def.ngModule`. The guard does not distinguish duplicate declarations from distinct provider-bearing wrappers. A bare module encountered before a wrapper can suppress its providers through the same guard.

This is source-confirmed at audit commit `4170bcc43`; the proposed regression has not yet been executed. The existing empty-wrapper duplicate test cannot expose provider loss.

## Expected fix

Preserve provider contributions and their order while avoiding duplicate declarations. Cover kept and mocked wrapper paths, separate tokens, later overrides of the same token, ordered multi providers, and bare-module/wrapper combinations. Keep exclude behavior and existing public signatures unchanged.

satanTime:
## Why

`ngMocks.guts` treats a module it has already visited as a duplicate even when a later `ModuleWithProviders` entry contributes different providers. This drops later registrations and overrides, including wrappers encountered after a bare module or through a later module's imports.

Fixes #15000.

## What

Retain provider-bearing wrappers in traversal order while continuing to omit empty duplicates. For a module that `guts` has already unpacked into TestBed, process the later providers through the existing provider resolver so its declarations stay in the same scope.

Regression cases cover separate tokens, overrides, ordered multi providers, kept and mocked modules, both bare-module orders, and wrappers reached through imports. They also check exclusions, provider-array preservation, private declaration visibility, and that mocked constructors and factories are not invoked.

## Impact

Repeated module configurations retain their providers without duplicating declaration ownership. Existing keep, exclude, and provider-mocking behavior remains in effect. Public APIs and documentation are unchanged because this restores the existing contract.


satanTime:

PR #15013, opened 2026-09-13
fix(providers): retain multi semantics for mocked declarations #15001

### LINKED ISSUE #15001: Mocked declaration providers lose multi-provider semantics
## Problem

Mocking an Angular declaration can change multi-provider injection from an array to a single instance. Two resolution paths lose the incoming `multi: true`: a previously cached declaration used as a provider, and a declaration's own `useExisting` alias.

This is a wider provider-resolution case reached while reviewing #1507 and #726. The earlier #121 / #124 fix explicitly preserved multi-provider metadata on cached provider objects, but the current cached-declaration branch bypasses that preservation. Reviewing #145 and the ordinary declaration-alias path also exposed the second case.

## Cached declaration reproducer

```ts
import { NgModule, Pipe } from '@angular/core';
import { MockModule } from 'ng-mocks';

@Pipe({ name: 'target', standalone: false })
class TargetPipe {
  public transform(value: string): string {
    return value;
  }
}

@NgModule({
  declarations: [TargetPipe],
  providers: [{ provide: TargetPipe, useClass: TargetPipe, multi: true }],
})
class TargetModule {}

it('preserves multi-provider shape for a cached declaration', () => {
  TestBed.configureTestingModule({ imports: [MockModule(TargetModule)] });
  expect(Array.isArray(TestBed.inject(TargetPipe))).toBe(true);
});
```

Native Angular with `imports: [TargetModule]` provides an array. The mocked module resolves the pipe declaration first, then produces a provider of the form `{ provide: TargetPipe, useClass: MockOfTargetPipe }`, without `multi`. Injection therefore has the wrong shape. With additional entries, this can also change aggregation or produce mixed-provider errors.

## Self-alias reproducer

```ts
import { Component, forwardRef, InjectionToken } from '@angular/core';
import { MockComponent } from 'ng-mocks';

const TOKEN = new InjectionToken<TargetComponent[]>('multi self alias');

@Component({
  selector: 'multi-alias-target',
  standalone: false,
  template: '',
  providers: [{
    provide: TOKEN,
    useExisting: forwardRef(() => TargetComponent),
    multi: true,
  }],
})
class TargetComponent {}

it('retains an array of aliased mock instances', () => {
  const mock = MockComponent(TargetComponent);
  TestBed.configureTestingModule({ declarations: [mock] });
  const fixture = TestBed.createComponent(mock);
  expect(fixture.debugElement.injector.get(TOKEN))
    .toEqual([fixture.componentInstance]);
});
```

Native Angular provides an array containing the component. The mock's rewritten alias omits `multi`, so it resolves to the component instance directly. Component `viewProviders` and directive `providers` share this branch.

## Cause and scope

- [`mock-ng-def`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-module/mock-ng-def.ts) processes declarations/imports before providers.
- [`create-resolvers`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-module/create-resolvers.ts) caches the resolved declaration in the shared definition stack.
- [`helper.resolve-provider`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/helper.resolve-provider.ts) returns `createFromResolution(...)` early on a cache hit. Only its non-cache path reattaches the parsed `multi` flag.

- [`clone-providers`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock/clone-providers.ts#L34-L46) rewrites an ordinary own alias with `toExistingProvider(provide, mockType)`. [`to-existing-provider`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock/to-existing-provider.ts) constructs only `provide` and `useExisting`; this path returns before the general provider resolver.

These are source-confirmed; the proposed Angular regressions have not yet been executed. It is separate from #14929's repeated module-wrapper collection and #14937's mixed-import ordering.

## Expected fix

Preserve multi-provider semantics through both existing paths. Pair focused source regressions with Angular module and declaration cases, including cached pipe/directive/component tokens, ordinary self aliases, providers/viewProviders callers, and ordered aggregation. Preserve alias identity, ordinary non-multi providers and existing keep/exclude behavior.

The special `NG_VALUE_ACCESSOR`, `NG_VALIDATORS`, and `NG_ASYNC_VALIDATORS` branches already construct lazy multi-provider proxies; preserve those factories and their current behavior. No new public option is needed.

satanTime:
## Why

Mocking a declaration that also acts as a multi provider could inject a scalar mock instead of an array. The cached declaration path returned before preserving `multi: true`. Aliases referring to their own component or directive lost the same flag when rewritten to the mock class.

## What

Preserve multi-provider metadata for cached resolutions and self aliases, including component `viewProviders`. Add regressions for components, directives and pipes, ordered aggregation, scalar aliases, configured provider identity, omitted providers and suppression of original constructors and factories.

## Impact

Consumers receive arrays containing the expected mock instances. Existing scalar-provider behavior, exclusions and configured recipes remain intact. This restores the existing provider contract and needs no public API or documentation changes.

Fixes #15001


satanTime:

PR #15014, opened 2026-09-13
fix(MockService): preserve nearest prototype member kinds #15002

### LINKED ISSUE #15002: MockService replaces child methods with inherited accessors
## Problem

`MockService` can turn a child class's callable method into an accessor returning `undefined` when a parent prototype has an accessor with the same name. Calling the mocked child method then throws.

This extends the accessor-mocking fix in #177. Memoizing decorators can represent a parent method as a runtime getter, as described in the older #109 discussion, while a subclass still overrides that method normally.

## Reproducer

```ts
import { MockService, ngMocks } from 'ng-mocks';

class BaseService {
  public read(): string {
    return 'base';
  }
}

class ChildService extends BaseService {
  public read(): string {
    return 'child';
  }
}

it('respects the nearest prototype descriptor', () => {
  // Represents the accessor descriptor installed by a memoizing method decorator.
  ngMocks.stubMember(BaseService.prototype, 'read', () => () => 'base', 'get');

  const mock = MockService(ChildService);
  expect(typeof mock.read).toBe('function');
  expect(mock.read()).toBeUndefined();
});
```

Expected: JavaScript resolves the child method before the parent's accessor, and the mock retains a callable dummy method. Current extraction records `read` as both a method and an accessor. Mock creation installs the method first, then overwrites it with a dummy getter/setter; reading `mock.read` returns `undefined`.

## Cause and scope

[`helper.extract-methods-from-prototype`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/helper.extract-methods-from-prototype.ts) tracks methods and accessors in separate sets. A name already found as a child method does not prevent a parent accessor from being collected. [`helper.create-mock-from-prototype`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/helper.create-mock-from-prototype.ts) then applies both categories in order.

This is source-confirmed at audit commit `4170bcc43`; the proposed regression has not yet been executed. The own-instance accessor case covered by #14916 is the reverse direction and does not prevent this prototype-shadowing failure. #14924 addresses accessor state replay rather than descriptor precedence.

## Expected fix

Respect the nearest descriptor across member kinds, following normal prototype lookup. Cover child method over parent accessor, the reverse arrangement, and deeper inheritance; preserve dummy callability, accessor behavior, and the absence of real getter side effects. Review shared MockProvider and wrapper-member discovery callers without adding public parameters or invoking constructors.

satanTime:
## Why

A child method can become an undefined accessor in `MockService` or `MockProvider` when an ancestor defines the same name as a getter. In the reverse direction, a mocked Angular component can receive a method instead of its child accessor. Prototype discovery collected the same key in both categories, allowing later mock initialization to select the wrong descriptor kind.

## What

Classify each key from its nearest prototype descriptor before inspecting ancestors. Regression cases exercise both inheritance directions, intermediate prototypes, symbol identity, declaration metadata seeds, and descriptor copying. Public cases check callable service mocks, provider identity, component bindings, accessor customization, and the absence of original constructor or accessor calls.

## Impact

Mocks retain the effective method or accessor declared by the original prototype chain. Existing APIs and defaults remain unchanged. Public documentation needs no update because this restores the existing mocking contract.

Fixes #15002


satanTime:

PR #15015, opened 2026-09-13
fix(MockService): mock objects using their actual prototypes #15003

### LINKED ISSUE #15003: MockService crashes on object shapes without a usable constructor
## Problem

The existing object-shape form of `MockService` assumes that the supplied object's public `constructor` property points to a constructor with a prototype. A legal application object with `constructor: undefined`, or a dictionary created with `Object.create(null)`, throws before its methods can be mocked. A getter named `constructor` is also executed during discovery.

This is a wider object-mocking case from #72 / #88. The original implementation guarded the constructor lookup; the current path no longer does. It is separate from #14916's own-accessor enumeration fix.

## Reproducer

```ts
import { MockService } from 'ng-mocks';

it('mocks an object whose constructor member is application data', () => {
  const source = {
    constructor: undefined,
    request: () => 'real',
  };

  const mock = MockService<typeof source>(source);
  expect(mock.request()).toBeUndefined();
  expect(source.request()).toBe('real');
});

it('mocks a null-prototype dictionary', () => {
  const source: { request: () => string } = Object.create(null);
  source.request = () => 'real';

  const mock = MockService<typeof source>(source);
  expect(mock.request()).toBeUndefined();
  expect(source.request()).toBe('real');
});
```

Expected: supplied object shapes have dummy methods without running real getters or constructors. Current source throws while reading `service.constructor.prototype`. The same input can occur as a nested object or an object-valued `useValue` provider.

## Cause and scope

- [The object handler](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/mock-service.ts#L33-L50) selects the prototype from `service.constructor.prototype` before inspecting own descriptors. The property can be absent, application data, or an accessor.
- [The object classifier](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/check.is-inst.ts) accepts null-prototype objects.
- [Descriptor lookup](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/helper.extract-property-descriptor.ts) also skips inspecting an object whose prototype is null. Simply guarding the constructor lookup will therefore leave another failure in the same object-mocking path.
- [Provider mocking](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-service/mock-provider.ts) delegates object-valued `useValue` providers to `MockService`.

The [documented object-shape API](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/docs/articles/api/MockService.md) does not require an inherited, callable `constructor` property.

This is source-confirmed; these proposed regressions have not yet been executed.

## Expected fix

Mock supplied objects using their actual prototype and descriptors, without reading an arbitrary `constructor` getter. Cover absent and non-constructor members, null-prototype shapes, nested shapes, and object-valued providers. Preserve ordinary class-instance prototype identity, accessor safety, method alias identity, configured spies, and the source object. Use the existing API without new public parameters.

satanTime:
## Why

Mocking a supplied object can throw or leave inherited methods real when its `constructor` property is missing, shadowed by application data, or exposed through a getter. Member discovery also skips user prototypes whose parent is null, and naming can invoke constructor getters.

## What

Use the supplied object's actual prototype and inspect constructor descriptors when naming mocks. Include terminal user prototypes in member and descriptor discovery while preserving the exclusion of built-in `Object.prototype`, including across browser realms.

Add regressions for constructor data and getters, null and custom prototypes, complete accessor and symbol descriptors, shared references and cycles, and object-valued module providers. Preserve native provider behavior and distinguish application prototypes from built-in prototypes without invoking original methods or getters.

## Impact

Object shapes and their providers produce usable mocks with the original prototype relationships and existing customization behavior. The documented public contract remains accurate, so public docs are unchanged.

Fixes #15003


satanTime:

PR #15016, opened 2026-09-13
fix(MockInstance): ignore absent optional initializers #15008

### LINKED ISSUE #15008: MockInstance registers absent optional initializers as callbacks
`MockInstance` accepts an optional initializer or configuration object, and the configuration's `init` property is optional. Passing `undefined`, `{}` or `{ init: undefined }` therefore satisfies the existing public signatures. These forms currently register an undefined callback and cause mock initialization to fail.

This is separate from #14997: no member name is supplied here. The valid-member-key correction leaves this existing initializer path unchanged.

### Reproduction

```ts
import { MockInstance, MockProvider, ngMocks } from 'ng-mocks';

@Injectable()
class TargetService {}

describe('optional MockInstance initializer', () => {
  MockInstance.scope();

  beforeEach(() =>
      providers: [MockProvider(TargetService)],
    }),
  );

  it('accepts an explicitly absent initializer', () => {
    MockInstance(TargetService, undefined);
    expect(ngMocks.get(TargetService)).toBeDefined();
  });

  it('accepts a configuration without an initializer', () => {
    MockInstance(TargetService, {});
    expect(ngMocks.get(TargetService)).toBeDefined();
  });
});
```

Replacing `{}` with `{ init: undefined }` reaches the same path. Each case uses fresh configuration, so the reproduction does not assume whether an empty configuration should clear or preserve earlier customizations.

### Current mechanism

`parseMockInstanceArgs` produces no key and no callback. Because an argument was supplied, `MockInstance` still registers that pair. `mock-instance-apply.ts` appends the undefined callback, which the service-provider and generated-declaration initialization paths then invoke.

This behavior was introduced by [8900fc37c](https://github.com/help-me-mom/ng-mocks/commit/8900fc37c6e6898a12908a8b00d73e01678f77fd), the fix for #2087: checking argument count correctly retained falsy member values, but also allowed absent optional initializers into the callback list. Before that change, these empty forms did not execute invalid callbacks.

The mechanism is confirmed by source review; this reproduction has not yet been executed.

### Expected scope

- Existing optional initializer/configuration forms must not lead to invalid callback execution.
- Preserve valid callback identity, arguments, returned-shape behavior and scope restoration.
- Preserve the documented no-argument reset and explicit member values, including undefined, false, zero and empty string.
- Establish empty-configuration behavior against the existing contract and history; do not assume all empty forms mean reset.
- No new public parameters, modes or live-instance feature are needed. PR #14818 retains this path and does not already address it.

The existing callback and reset documentation is in [MockInstance](https://ng-mocks.sudo.eu/api/MockInstance). The historical reset report #1046 concerns `MockInstance(Target)` with the second argument omitted; it does not settle the semantics of `{}`.

satanTime:
## Why

`MockInstance(Target, undefined)`, `MockInstance(Target, {})` and `MockInstance(Target, { init: undefined })` registered an undefined callback. Mock providers and generated Angular declarations then failed when they tried to invoke it.

## What

Skip registration when neither a member key nor an initializer is supplied. Add source registration and service, component and token regressions covering all three optional forms, callback identity/order/results, scope restoration and reset behavior.

## Impact

Supplied empty optional configurations are no-ops that preserve prior customization. Calling `MockInstance(Target)` without a second argument still resets customization. Explicit member values, including undefined and other falsy values, retain their existing behavior. No public API change is needed.

Fixes #15008


satanTime:

PR #15017, opened 2026-09-13
fix: respect child query and input metadata #15006

### LINKED ISSUE #15006: Child declarations retain inherited query selectors and input transforms
## Problem

A child component or directive that redeclares an inherited property can be mocked with the parent's metadata instead of its own:

- A child content query keeps the parent's selector.
- A child decorator input keeps the parent's transform for the same public binding.

The real Angular declaration uses the child's metadata in these cases, so the mock can select the wrong projected element or assign a value of the wrong type. These are wider metadata-inheritance cases from #2, #28, and #69.

## Content-query reproducer

```ts
import { Component, ContentChild, Directive, ElementRef } from '@angular/core';
import { MockComponent, ngMocks } from 'ng-mocks';

@Directive({ standalone: false })
class Base {
  @ContentChild('first')
  public item: ElementRef<HTMLElement> | undefined = undefined;
}

@Component({
  selector: 'query-child',
  standalone: false,
  template: '<ng-content></ng-content>',
})
class Child extends Base {
  @ContentChild('second')
  public override item: ElementRef<HTMLElement> | undefined = undefined;
}

@Component({
  standalone: false,
  template: '<query-child><span #first>first</span><span #second>second</span></query-child>',
})
class Host {}

it('uses the child content query on the mocked dependency', () => {
  TestBed.configureTestingModule({ declarations: [Host, MockComponent(Child)] });
  const fixture = TestBed.createComponent(Host);
  fixture.detectChanges();
  expect(ngMocks.findInstance(Child).item?.nativeElement.textContent).toBe('second');
});
```

A native control using `Child` selects the second span. Current mock metadata retains `ContentChild('first')`, so it selects the first span. The same property-key guard affects plural content queries and query options.

## Input-transform reproducer

This example uses decorator input transforms available from Angular 16.1.

```ts
import { Component, Directive, Input } from '@angular/core';
import { MockComponent, ngMocks } from 'ng-mocks';

@Directive({ standalone: false })
class Base {
  @Input({ transform: Number }) public value: number | string = 0;
}

@Component({ selector: 'transform-child', standalone: false, template: '' })
class Child extends Base {
  @Input({ transform: String }) public override value: number | string = '';
}

@Component({
  standalone: false,
  template: '<transform-child [value]="3"></transform-child>',
})
class Host {}

it('uses the child input transform on the mocked dependency', () => {
  TestBed.configureTestingModule({ declarations: [Host, MockComponent(Child)] });
  TestBed.createComponent(Host).detectChanges();
  expect(ngMocks.findInstance(Child).value).toBe('3');
});
```

A native control using `Child` applies `String` and stores `'3'`. Current mock metadata applies the inherited `Number` and stores numeric `3`.

## Cause and scope

At audit commit `4170bcc430d95e001c92309cb12f7c7dfb043dc1`, [`collect-declarations`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/resolve/collect-declarations.ts) parses the parent and copies its input/query entries before reading the child.

The single/plural query parsers and compiled property-decorator parser only assign a query when that member key is absent. The copied parent entry prevents a child replacement. The input merger similarly retains `defTransform ?? transform` for a same-name/same-alias binding; later effective compiled or reflected child metadata passes through the same merger and cannot correct the inherited transform.

The resulting metadata is applied to mocks by [input decoration](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/common/decorate.inputs.ts) and [query decoration](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/common/decorate.queries.ts). Angular's native inheritance path executes parent content queries before child queries, and retains a child's own input definition over its parent's.

These failures are source-confirmed against current ng-mocks and Angular inheritance code. The proposed native/mock regressions have not yet been executed. This is separate from #14918's inherited structural alias case and #14919/#14920's signal metadata and query policies.

## Expected fix

Honor effective child metadata through the existing collector. Cover same-member ContentChild/ContentChildren redeclarations and same-binding decorator input transform replacement, with native controls. Check removal of an inherited transform and child query options where the native control establishes the expected result.

Preserve inherited members that the child does not redeclare, distinct input aliases, and established precedence between multiple metadata sources within one class. Do not require a mock to restore its original component view or change intentional signal-query omission. No new public API or parameters are needed.

satanTime:
## Why

A child component or directive could retain its parent's content-query selector or input transform after redeclaring the same property. The collector loaded inherited entries before the child's metadata, so its normal duplicate handling kept the parent entry.

## What

Collect child queries and inputs before combining them with inherited metadata. Replace the effective input transform, including removal, while retaining inherited bindings and their alias application order. Respect explicitly redeclared input options while preserving required metadata when a compiled definition only repeats an inherited binding. Preserve existing precedence between metadata sources within one class.

Add native and mocked component/directive regressions for inherited content queries and input transforms, plus collector cases for all four query decorators, untouched inherited metadata, explicit overrides, and parent/sibling cache isolation. Keep transform replacement and removal at their respective Angular compatibility boundaries.

## Impact

Mocks use the child declaration's query selectors and effective input transforms while unrelated inherited metadata remains available. Existing signal-query behavior and mocked lifecycle suppression are preserved. Public documentation remains accurate and does not need changes.

Fixes #15006


satanTime:

PR #15018, opened 2026-09-13
fix(MockBuilder): propagate asynchronous compilation failures #15004

### LINKED ISSUE #15004: MockBuilder leaves its promise pending when Angular compilation rejects
## Problem

When `TestBed.compileComponents()` rejects asynchronously, the promise returned by `MockBuilder` never settles. Returning the builder from `beforeEach`, awaiting it, or attaching `.catch()` can produce a timeout or an unhandled rejection instead of reporting the original compilation error.

This was found while reviewing the MockBuilder workflow introduced by #70 / #91 and its shared compilation path.

## Reproducer

A public Angular resource loader can make compilation reject without patching compiler internals. Start with an inline component and add an external style in the documented pre-compilation hook so building the initial declaration does not depend on unresolved resource metadata.

```ts
import { Component } from '@angular/core';
import { ResourceLoader } from '@angular/compiler';
import { MockBuilder } from 'ng-mocks';

@Component({ selector: 'compile-target', standalone: false, template: '' })
class TargetComponent {}

it('rejects with the original Angular compilation error', async () => {
  const originalError = new Error('stylesheet load failed');
  const pending = MockBuilder()
    .keep(TargetComponent)
    .beforeCompileComponents(testBed => {
      testBed.configureCompiler({
        providers: [{
          provide: ResourceLoader,
          useValue: { get: () => Promise.reject(originalError) },
        }],
      });
      testBed.overrideComponent(TargetComponent, {
        set: { styleUrls: ['missing.css'] },
      });
    })
    .then();

  await expectAsync(pending).toBeRejectedWith(originalError);
});
```

Expected: rejection reaches the caller with the original error. Current source leaves `pending` unresolved. A native `TestBed.compileComponents()` setup with the same resource loader should be the control when implementing the regression.

## Cause and scope

At audit commit `4170bcc430d95e001c92309cb12f7c7dfb043dc1`, [`MockBuilderPromise.then`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock-builder/mock-builder.promise.ts#L220-L236) creates an outer Promise with only a resolve callback. It calls `compileComponents().then(success)` without connecting rejection to the outer Promise and discards the derived promise.

The supplied rejection callback and the existing `.catch()` / `.finally()` methods all operate on the pending outer promise. Synchronous exceptions during build, configuration, or the pre-compilation hook already reject through the Promise executor; they do not cover this asynchronous branch.

The source failure is definite. The proposed Angular resource-loader reproducer was checked against Angular's compilation source but has not yet been executed.

## Expected fix

Propagate asynchronous compilation failures through the existing thenable and preserve error identity. Cover native asynchronous rejection as well as successful compilation and synchronous failures. Keep the existing result shape, callback behavior, and public API.

satanTime:
## Why

When Angular rejects `compileComponents()`, MockBuilder's outer promise remains pending. Awaiting the builder or returning it from setup can time out instead of reporting the compilation error. This affects the shared promise path introduced in #91.

## What

Forward compilation rejections to the existing builder promise without changing synchronous TestBed configuration, pre-compilation callbacks, or the successful result. Cover real ResourceLoader failures and successful compilation, ordinary recovery after TestBed reset, and the existing `then`, `catch`, and `finally` paths. A portable compiler-rejection regression remains active for every supported version. The real resource cases exclude Angular 9–16 because their packaged Ivy TestBed compiler uses a resource queue separate from Angular core.

## Impact

Callers receive the original compilation failure and can handle it through the existing promise API. No new options or entry points are needed; public documentation remains unchanged because this restores the existing contract.

Fixes #15004


satanTime:

PR #15019, opened 2026-09-13
fix: initialize outputs on host-bound members #15007

### LINKED ISSUE #15007: HostBinding on an output prevents the mocked EventEmitter from being initialized
## Problem

A property decorated with both `@HostBinding` and `@Output` is collected correctly but becomes an unusable output on a mocked component or directive. Binding the output can fail because its mock value is undefined instead of an EventEmitter.

This is a wider case of #5's multiple-property-decorator fix. Existing HostBinding + Input coverage does not exercise output initialization.

## Reproducer

```ts
import { Component, EventEmitter, HostBinding, Output } from '@angular/core';
import { MockComponent, ngMocks } from 'ng-mocks';

@Component({ selector: 'event-child', standalone: false, template: '' })
class Child {
  @HostBinding('attr.data-event')
  @Output('changed')
  public readonly event = new EventEmitter<string>();
}

@Component({
  standalone: false,
  template: '<event-child (changed)="value = $event"></event-child>',
})
class Host {
  public value = '';
}

it('keeps the output usable on the mocked dependency', () => {
  TestBed.configureTestingModule({ declarations: [Host, MockComponent(Child)] });
  const fixture = TestBed.createComponent(Host);
  fixture.detectChanges();
  ngMocks.findInstance(Child).event.emit('sent');
  expect(fixture.componentInstance.value).toBe('sent');
});
```

A native control using `Child` can subscribe and receive the emitted value. Current source predicts that rendering the mock fails while Angular subscribes to the undefined output. The output contract is the relevant behavior; mock host bindings are intentionally suppressed.

## Cause and scope

- [Declaration decoration](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/mock/decorate-declaration.ts) records both the explicit output and the host-binding member.
- [The shared Mock constructor](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/common/mock.ts#L224-L253) first installs dummy getter/setter descriptors for host bindings, then initializes inputs and outputs.
- [`applyOutputs`](https://github.com/help-me-mom/ng-mocks/blob/4170bcc430d95e001c92309cb12f7c7dfb043dc1/libs/ng-mocks/src/lib/common/mock.ts#L119-L131) skips a member whenever an own descriptor already exists, even when that descriptor was just generated for a host binding and returns undefined. It therefore never installs the output EventEmitter.

This affects the shared component/directive initialization path, including aliased and unaliased outputs. It is source-confirmed; the proposed Angular regression has not yet been executed.

## Expected fix

Initialize a declared output despite the library's generated host-binding storage. Pair a focused initialization test with real/mocked Angular component and directive cases that subscribe and emit. Preserve aliases, instance-local emitters, original constructor/getter suppression, explicit customization, and ordinary HostBinding + Input behavior. No real host-binding DOM effects, signal expansion, or new public parameters are required.

satanTime:
## Why

When a component or directive member has both `@HostBinding` and `@Output`, generated host-binding accessors cause the output initializer to skip it. Angular then fails to subscribe because the mock output is undefined. This also affects inherited outputs and compiler-declared `output()` members.

## What

Skip generated host-binding storage for members already declared as outputs, allowing the existing EventEmitter initialization to run. Cover real and mocked component/directive subscriptions, aliased decorator and `output()` declarations, inherited getter-backed outputs, independent default emitters, and explicit emitter customization. Preserve ordinary host-bound input updates and the suppression of original constructor, getter, and host DOM effects.

## Impact

Mocked outputs remain subscribable when their backing members are also host-bound. Existing custom output accessors and explicitly supplied emitters retain their behavior. This restores the documented output contract, so public documentation is unchanged.

Fixes #15007


satanTime:

PR #15021, opened 2026-09-13
ci: fix diagnostics in regression tests and signature verification

satanTime:
## Why

Successful CI runs still print component identity and collection tracking warnings from regression fixtures. Tests of inherited injectable definitions and failed asynchronous compilation also print the native diagnostics they deliberately exercise. Codecov signature verification creates a keybox that causes a GnuPG lock diagnostic.

## What

Give the selectorless override fixtures distinct host attributes and use stable position tracking for the fixed single-row signal-form fixture. Guard the affected suites with `ngMocks.throwOnConsole`. Isolate intentional inherited-DI warnings and asynchronous compilation errors with scoped `ngMocks.ignoreOnConsole` spies, check every recorded diagnostic, and preserve the original construction, rejection, and recovery assertions. The shared helpers restore console methods after each test.

Decode the downloaded Codecov public key into a binary OpenPGP keyring before verification, retaining the signature check, checksum check, and verification-before-execution order.

## Impact

The regression cases retain their existing coverage with clearer output. Signature verification uses a binary OpenPGP keyring from the same key source and release artifacts. Library behavior and public APIs are unchanged.


satanTime:

PR #15023, opened 2026-09-14
build: pin legacy browser targets to Angular release versions

satanTime:
## Why

Angular 6–9 use floating browser queries even though their projects exercise historical Angular toolchains. Browser-data updates can therefore change their compilation targets. Angular 6 and 7 also keep the configuration below the directory where ng-packagr starts its lookup.

## What

Replace the floating queries with fixed Chrome, Firefox, ESR, Edge, Safari, and iOS version ranges based on each original Angular major release. Preserve the existing IE exclusion for Angular 6–8 and inclusion for Angular 9. Move the Angular 6 and 7 configurations to their project roots so application styles and library builds can find the same targets.

## Impact

Legacy compilation targets no longer advance with new browser releases. Existing compatibility-data groupings still apply, including grouped iOS minor versions. The installed Browserslist data-age warning remains because its age check runs independently of query selection.

Browser binaries, library behavior, and public APIs are unchanged. Public documentation changes are unnecessary because this only configures the versioned development projects.

Follow-up to #15021.


satanTime:

PR #15025, opened 2026-09-14
Fix/collect declarations cache poisoning on throw

### LINKED ISSUE #15024: Bug: parse() permanently poisons a def's metadata cache when a step throws mid-parse
## Problem

`collectDeclarations()` caches a class's parsed metadata directly on the class object itself, using two properties: `__ngMocksParsed` (a "done" flag) and `__ngMocksDeclarations` (the result). The flag is written before the result. If anything in the middle of parsing a given `def` throws (for example `ReflectionCapabilities.parameters()` on an edge case class), the `def` ends up with `__ngMocksParsed = true` but `__ngMocksDeclarations` never assigned.

Since production classes are shared objects across every spec file in a suite (loaded once, not per spec, under bundlers that split each spec into its own entry while sharing a common chunk for application code), any later, unrelated spec that touches that same class hits the early return path and gets back `undefined`, or worse, whatever unrelated declarations happen to already be cached higher up its prototype chain. Depending on which of those two happens, the effect is either a thrown `TypeError: Cannot read properties of undefined (reading 'length')` inside `getNgType`, or a silently wrong result: the affected class looks like it has no decorators, inputs, or outputs at all, with nothing raised anywhere.

In a suite large enough to have many shared classes, this makes an unrelated, seemingly random spec fail intermittently, always with the same stack trace but a different failing spec each run, since it depends on whichever class happens to hit the original throw first and whichever later spec happens to touch it again.

## Reproducer

Dropped into `collect-declarations.spec.ts`, next to the other tests using the same `global.__ngMocksReflectComponentType` override:

```ts
it('does not permanently cache a def as parsed when a parsing step throws', () => {
  const global = funcGetGlobal();
  const def: any = {
    ɵcmp: {
      inputs: {
        value: 'value',
      },
    },
  };

  global.__ngMocksReflectComponentType = () => {
    throw new Error('boom');
  };
  expect(() => collectDeclarations(def)).toThrow();

  global.__ngMocksReflectComponentType = false;
  const actual = collectDeclarations(def);
  expect(actual.inputs).toEqual(['value']);

  delete global.__ngMocksReflectComponentType;
});
```

Current source returns `[]` for `actual.inputs` on the second call, instead of `['value']`. The first, forced throw leaves `def` marked as parsed, so the second call never reparses it and falls back to whatever was cached on `Object.prototype` from an earlier, unrelated call.

We also reproduced the same mechanism against a real, unmodified production component from our own application, using the published `ng-mocks@14.14.0` package as installed via npm, with no rebuild and no forked version involved. That test, plus a captured run without the fix and a captured run with the fix applied directly to the installed package, are attached (`notifications.ngmocks-cache-poisoning.diag.spec.ts`, `without-fix-error-test-simulation.txt`, `with-fix-error-test-simulation.txt`). The original crash (`TypeError: Cannot read properties of undefined (reading 'length')`) is attached as a screenshot as well:

Image

## Cause and scope

At commit `39b879ede4dea30e2b28ae84a9dca8b780e34ad8`:

* [`parse()`](https://github.com/help-me-mom/ng-mocks/blob/39b879ede4dea30e2b28ae84a9dca8b780e34ad8/libs/ng-mocks/src/lib/resolve/collect-declarations.ts#L622-L686) writes `coreDefineProperty(def, '__ngMocksParsed', true)` at line 634, before any of the parsing steps that follow it (`parseParameters`, `parseAnnotations`, `parseNgDef`, `parseReflectComponentType`, and finally `reflectionCapabilities.parameters(def)` at line 682, which is where `__ngMocksDeclarations` actually gets written).
* If any of those steps throws, execution never reaches line 679's `coreDefineProperty(def, '__ngMocksDeclarations', ...)`, but the flag from line 634 is already set and stays that way for the lifetime of the class object.
* The early return at the top of the same function, `if (hasOwnProperty(def, '__ngMocksParsed')) return def.__ngMocksDeclarations;`, is what turns this into a permanent, silent failure for every later caller instead of a one time, localized one.

This has been verified against the actual `main` source above, against ng-mocks' own test suite (a regression test reproducing it is included in the fix branch), and against a real production class from our project (editing minified node modules code, as below, to prove the error), both without and with the fix applied.

Image

## Expected fix

A `def` that fails to parse should be left exactly as if `parse()` had never touched it, so a later call can simply retry, instead of being poisoned for the rest of the run. Both properties should only be written together, once parsing has actually succeeded, with no change to the result on the success path.

## Environment

* `ng-mocks`: 14.14.0
* Angular: 20.x (esbuild based `@angular/build:karma` builder, Karma + ChromeHeadless)
* Suite size: about 11,000 specs in one `ng test` run

## Attachments

[notifications.ngmocks-cache-poisoning.diag.spec.txt](https://github.com/user-attachments/files/32202468/notifications.ngmocks-cache-poisoning.diag.spec.txt) (spec as txt so i could upload to github)
[without-fix-error-test-simulation.txt](https://github.com/user-attachments/files/32202417/without-fix-error-test-simulation.txt)
[with-fix-error-test-simulation.txt](https://github.com/user-attachments/files/32202446/with-fix-error-test-simulation.txt)

mayhmemo:
## Why

`parse()` caches a def's parsed metadata directly on the class object, writing the "already parsed" flag (`__ngMocksParsed`) before the parsed result (`__ngMocksDeclarations`) is actually built. If any step in between throws for a given def, that def is left permanently flagged as parsed with no declarations ever stored.

Production classes are shared objects across every spec file in a suite, not copies per spec, under bundlers that split each spec into its own entry while sharing a common chunk for application code. A def poisoned this way stays broken for the rest of the run: any later, unrelated spec that touches the same class hits the early return path and gets back undefined, or whatever unrelated declarations happen to already be cached higher up its prototype chain. The visible symptom is a TypeError inside `getNgType`, or a silently wrong result where the affected class looks like it has no decorators, inputs, or outputs at all.

## What

Build the full declarations object first, then write `__ngMocksParsed` and `__ngMocksDeclarations` together, only once parsing has actually succeeded. A def that fails to parse is left exactly as it was before `parse()` touched it, so a later call can retry instead of inheriting a permanently broken cache.

Add a regression test that forces a throw partway through parsing a def, asserts the throw happens, then retries the same def and asserts it parses correctly this time, catching the exact silent fallback to inherited Object.prototype declarations that the bug produces.

## Impact

Behavior on the success path is unchanged: every existing test in the suite passes, along with lint, prettier, and ts:check. A def that fails to parse for any reason no longer poisons the shared cache for every other spec that happens to touch the same class later in the run. The fix does not add any public API, switch, or parameter.

Fixes #15024


satanTime:

PR #15035, opened 2026-09-14
fix(ngMocks.change): recognize _controlValueAccessorChangeFn #15030

### LINKED ISSUE #15030: fix(ngMocks.change): recognize _controlValueAccessorChangeFn on real CVAs
## Problem

The default `ngMocks.change(host, value)` call cannot drive a real ControlValueAccessor whose registered callback is stored as `_controlValueAccessorChangeFn`. Angular Material's checkbox, radio group and button-toggle group use this name. Their value-accessor registration and injector identities are correct, but the helper throws its unsupported-ControlValueAccessor error.

The failure was observed with real Material and with Material kept through MockBuilder on Angular/forms/Material 22.1.6, using ng-mocks at `f63b4033f7ef4195550ebfcc537df0be1c4e8a6c`. Mocked versions work through their forwarding proxy. The explicit third argument also works:

```ts
// Existing workaround after finding the control host.
ngMocks.change(input, 'updated', '_controlValueAccessorChangeFn');

// Expected simple usage after this change.
ngMocks.change(input, 'updated');
```

This is a callback-discovery limitation, not evidence that Angular failed to connect the real CVA. Material group values belong to the group host, not its individual buttons. Boolean checkbox values and group option values must retain their respective meanings.

## Regression to write without Material

Create local declarations under `tests/issue-15030/test.spec.ts`: an own component CVA, an `NG_VALUE_ACCESSOR` multi-provider using `useExisting`, and a parent using classic forms. The child should have prototype `writeValue`, `registerOnChange`, `registerOnTouched`, and `setDisabledState` methods; `registerOnChange` stores its callback in `_controlValueAccessorChangeFn`. Give it a plain custom-element host with no input/change listeners so the reproducer reaches the callback path. Include an own UI token alias if needed to assert the rendered instance independently of the CVA.

Add `signals.spec.ts` with the same contract under real FormField. Keep required root providers. Compare plain TestBed, MockBuilder with the own child kept, and the own child mocked. **Do not import or install `@angular/material` for these regressions or spread targets.** Existing Material integration dependencies may corroborate the investigation, but are not the reproducer or a prerequisite for fixing it.

## Acceptance criteria

- [ ] The own real/kept CVA fails with the default helper before the fix and passes afterward; the mocked version stays working.
- [ ] Default and explicit-name calls update the intended parent exactly once; initial/later parent values reach the same child.
- [ ] Default user edits mark the applicable real form dirty; separate touch works without changing the value. Disabled-state forwarding and another control's value/state remain intact.
- [ ] Original class and UI token resolve to the rendered child. A mocked accessor may be a proxy, but must forward to that exact child with registered callbacks.
- [ ] Native listener routing, currently recognized callback names, and explicit method-name behavior are preserved. Do not globally move guessed callback names ahead of native events.
- [ ] Classic declarations use APIs compatible with Angular 5–22. Gate the separate signal file with `features=signalForms`; use the repository's actual View Engine/Ivy and zoned/zoneless profiles. Pair changed callback-discovery code with a focused source-unit case.
- [ ] Keep setup, actions and assertions inline, with simple comments explaining what and why. Use static imports and per-file spread gates; no test helpers or runtime skips.

## Starting points and scope

- [Default callback names and dispatch](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/libs/ng-mocks/src/lib/mock-helper/cva/mock-helper.change.ts) and its adjacent unit spec.
- [Existing signal CVA regression](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/tests/issue-14909/test.spec.ts): real/mock callback setup.
- [Kept alias/value/state regression](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/tests/issue-14922/test.spec.ts): identity and single-emission assertions.
- Follow repository Docker wrappers, coverage and affected spread validation; keep final PR CI green. No new validators or Material UI behavior. Update the change API only if its documented public contract needs correction.

## Relationships

Related supported-host documentation: #14988; original guide request: #14925.

satanTime:
## Why

`ngMocks.change(host, value)` rejects valid real ControlValueAccessors that store Angular's registered callback as `_controlValueAccessorChangeFn`. Material's checkbox, radio group and button-toggle group use this name, so keeping those controls currently requires the explicit method-name argument.

## What

Recognize the callback name after the existing names. Add own-component regressions for real, kept and mocked CVAs with `ngModel`, reactive forms and signal forms, including injector identity, parent writes, single updates, touch, disabled-state forwarding and sibling isolation. Focused source tests preserve existing callback precedence, explicit overrides and native event routing.

## Impact

The default helper call works with these registered callbacks. Existing callback selection and mock dispatch remain intact. The regressions have no Material dependency; classic regressions use APIs compatible with the supported Angular range and signal forms use their feature gate. No public documentation change is needed.

Fixes #15030.
Related to #14909.


satanTime:

PR #15036, opened 2026-09-14
fix(forms): preserve mocked value accessor categories #15031

### LINKED ISSUE #15031: fix(forms): preserve default, built-in and custom accessor categories when mocking
## Problem

Mocking both a default Angular value accessor and a custom accessor on the same host can make a valid form fail to render. At ng-mocks revision `f63b4033f7ef4195550ebfcc537df0be1c4e8a6c`, with Angular/forms/Material **22.1.6**, real reactive forms plus mocked `DefaultValueAccessor` and mocked `MatAutocompleteTrigger` produce:

```text
More than one custom value accessor matches form control with unspecified name attribute
```

The equivalent real, kept, and Material-mocked configurations work when the default accessor stays real. This is a separate accessor-selection defect; it does not establish the cause of the unresolved MatSelect diagnostics in [the original report](https://github.com/help-me-mom/ng-mocks/issues/14909#issuecomment-5663196874).

`clone-providers.ts` replaces each `NG_VALUE_ACCESSOR` provider with a `MockControlValueAccessorProxy`. Angular distinguishes default, built-in, and custom accessors through constructor/prototype identity. Both proxies consequently appear custom, so Angular rejects an otherwise valid pair before the helper can change its value.

## Regression shape

Add small repository-owned declarations under `tests/issue-15031`. **Do not install or import Angular Material in core or spread regressions.** Material supplied the observed contract; the regression should express it with Angular and ng-mocks alone.

- A parent owns a `FormControl` and imports real `ReactiveFormsModule`.
- An `input[ownCva]` directive implements `ControlValueAccessor` using ordinary prototype methods and provides a multi `NG_VALUE_ACCESSOR` self alias.
- Give that directive a separate `OWN_CONTROL` injection-token self alias to check rendered-instance identity independently of its forwarding accessor.
- Render `<input ownCva [formControl]="control">`, which also matches Angular's `DefaultValueAccessor`.
- Compare plain TestBed, kept declarations, and mocked declarations, including both accessors mocked together while forms remain real.
- Use a number input without the own custom directive to verify that Angular's built-in `NumberValueAccessor` wins over the default accessor. Add the own custom directive to that host in a separate case to verify custom-over-built-in precedence. Add a second own custom directive to verify that two actual custom accessors remain invalid.

Keep setup and assertions inline, following the existing issue-spec style. Include the new issue link near the suite.

## Implementation boundary

Preserve Angular's accessor categories when constructing mock providers/proxies, while retaining forwarding to the correct rendered declaration. Do not select the first provider, suppress duplicate-accessor errors, add a Material-specific exception, or change Angular's selection implementation.

Relevant source: [provider cloning](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/libs/ng-mocks/src/lib/mock/clone-providers.ts), [CVA proxy](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/libs/ng-mocks/src/lib/common/mock-control-value-accessor-proxy.ts), and [proxy attachment](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/libs/ng-mocks/src/lib/common/mock.ts). Pair the spread regression with focused source-unit coverage for the category-preservation mechanism.

## Acceptance criteria

- [ ] Default/built-in/custom precedence matches Angular in real, kept, and mixed/mock candidate combinations. Provider order does not change the selected category.
- [ ] Two genuine custom candidates still raise Angular's duplicate-custom error, including when both are mocked.
- [ ] Local class and `OWN_CONTROL` lookups identify the rendered directive. `NG_VALUE_ACCESSOR` contains the selected candidate; the mock candidate's `instance` points to that directive rather than another host or a separate service.
- [ ] Initial and later parent values reach the selected declaration. A helper change updates only its parent once; separate touch works. The unselected accessor does not become the active connection.
- [ ] Classic fixtures use APIs compatible with Angular 5–22: Angular 5–8 `view-engine-zoned`; 9–11 both zoned engine profiles; 12–19 `ivy-zoned`; 20–22 both `ivy-zoned` and `ivy-zoneless`. Avoid unnecessary standalone or signal-form gates. Any additional newer-API fixture gets its actual version/feature gate in `test-spread.conf`.
- [ ] Use static imports, simple comments explaining what happens and why, inline setup/actions/assertions, and no test helpers or runtime skips. Validate through repository Docker wrappers, focused source coverage and affected spread profiles; keep the implementation PR CI green.

Existing Material integration suites are reference evidence only. The core regression and fix must not require their dependencies. No new validator behavior or Material UI emulation is in scope.

References: [issue-157 accessor combinations](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/tests/issue-157/test.spec.ts), [issue-305 mocked default accessor](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/tests-e2e/src/issue-305/test.spec.ts), and [issue-14922 kept accessor identity](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/tests/issue-14922/test.spec.ts).

satanTime:
## Why

Mocking Angular's default or built-in value accessors changes their constructor identity. Angular then classifies these proxies as custom accessors, so otherwise valid combinations can throw a duplicate-custom-accessor error or select the wrong accessor.

## What

Preserve the original declaration's constructor identity on each CVA proxy while retaining its proxy prototype and attachment target. Add repository-owned regressions for real, kept and mixed/mock accessor combinations, selected-instance identity, parent writes, helper updates and sibling isolation. Source coverage exercises Angular's selection order, built-in categories, custom subclasses, duplicate errors and callback forwarding.

## Impact

Mocked accessors retain Angular's custom-over-built-in-over-default precedence. Invalid custom combinations still fail, and proxies continue forwarding to the rendered declaration. No Material dependency or public API change is introduced; the fix restores the existing contract and requires no documentation update.

Fixes #15031.
Related to #14909.


satanTime:

Issue #15033, opened 2026-09-14
feat(forms): support change and touch through mocked FormField bindings


satanTime:
## Problem and intended behavior

Mocked NgModel and FormControlDirective have binding-level helper fallbacks, but a mocked FormField currently has no equivalent. With a mocked self-registering child CVA, an attached but unregistered accessor can make `ngMocks.change` silently do nothing. A native input with only mocked FormField instead reaches the required-accessor error. Neither connects the supplied real FieldTree to the requested edit.

Introduce an explicit binding-level contract for intentionally mocked FormField so the same find → read → change → assert flow works. This does not restore Material UI or all of Angular's forms lifecycle.

The disconnected child behavior was observed with Angular/forms/Material 22.1.6 at ng-mocks `f63b4033f7ef4195550ebfcc537df0be1c4e8a6c`. Existing [MockFormBindings:signals](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/examples/MockFormBindings/signals.spec.ts) explicitly documents and asserts the current unsupported native case. This issue adds capability; do not describe it as already supported or confuse it with the original reporter's still-unreproduced null-accessor setup.

## Regression to write without Material

Create own parent and child declarations under `tests/issue-15033/`. The parent owns `inputModel = signal({ inputValue: 'initial' })` and `inputForm = form(inputModel)`. Bind an own custom control with `[formField]="inputForm.inputValue"`.

The child should reproduce constructor registration through local optional NgControl, prototype CVA methods, an ordinary own UI token alias, and **no** NG_VALUE_ACCESSOR provider. Mock both the child and FormField, retaining required root services. Add a native input case and a separate signal-model child case, using actual feature gates for those APIs. Do not import/install Material, subclass Material declarations, or introduce a Material dependency into older spread targets.

```ts
// FormField and the own child are intentionally mocked by this setup.
beforeEach(() => MockBuilder(ParentComponent).keep(NG_MOCKS_ROOT_PROVIDERS));

// Inside the test, after rendering ParentComponent.
const input = ngMocks.reveal(['formField', component.inputForm.inputValue]);
expect(component.inputModel().inputValue).toBe('initial');

// The helper applies an edit to the supplied field binding.
ngMocks.change(input, 'updated');
expect(component.inputModel().inputValue).toBe('updated');
expect(component.inputForm.inputValue().dirty()).toBe(true);

[...]

PR #15037, opened 2026-09-14
fix(forms): use binding fallbacks for unregistered accessors #15032

### LINKED ISSUE #15032: fix(forms): prevent unregistered CVA proxies from masking mocked bindings
## Problem

At ng-mocks `f63b4033`, with Angular/forms/Material 22.1.6, mocking both MatSelect and its classic forms directives makes `ngMocks.change` return without updating the parent. The two classic cases reproduce in Chrome/Jasmine and Vitest. Keeping the forms infrastructure real works with real, kept, and mocked MatSelect.

The local `NgControl.valueAccessor` is a **non-null proxy attached to the correct rendered mock**. No forms callback was registered. This is different from the reporter's exact null-accessor diagnostics in [#14909](https://github.com/help-me-mom/ng-mocks/issues/14909#issuecomment-5663196874); those remain unreproduced.

[Accessor lookup](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/libs/ng-mocks/src/lib/mock-helper/cva/func.get-vca.ts) returns that proxy before considering the supplied `FormControl` or mocked `NgModel`. [Change dispatch](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/libs/ng-mocks/src/lib/mock-helper/cva/mock-helper.change.ts) treats the mock as connected and invokes its inherited no-op `__simulateChange`. Consequently, the existing `setValue` and `NgModel.update.emit` fallbacks are unreachable. Invoking those underlying operations directly updates the parent successfully.

## Own-component regression fixture

Add a small constructor-registering CVA under `tests/issue-15032/test.spec.ts`; no Material imports or dependency installation are required. The following Angular 22 fixture outline has not yet been executed; adapt executable compatibility syntax using the repository's existing classic examples. Its CVA methods must remain prototype methods, and its UI alias must resolve to the rendered declaration without introducing `NG_VALUE_ACCESSOR`:

```ts
const UI_CONTROL = new InjectionToken<unknown>('UI_CONTROL');

@Component({
  selector: 'own-select',
  standalone: false,
  template: '{{ value }}',
  providers: [{
    provide: UI_CONTROL,
    useExisting: forwardRef(() => OwnSelect),
  }],
})
class OwnSelect implements ControlValueAccessor {
  public value: unknown;
  public onChange = (value: unknown): void => {};
  public onTouched = (): void => {};

  constructor(@Optional() @Self() control: NgControl) {
    if (control) control.valueAccessor = this;
  }

  public writeValue(value: unknown): void { this.value = value; }
  public registerOnChange(fn: (value: unknown) => void): void {
    this.onChange = fn;
  }
  public registerOnTouched(fn: () => void): void {
    this.onTouched = fn;
  }
}
```

Declare the parent and child in a module importing `FormsModule` and `ReactiveFormsModule`. The parent has `inputValue = 'initial'`, `inputControl = new FormControl('initial')`, and these bindings:

```html
<own-select name="inputName" [(ngModel)]="inputValue"></own-select>
<own-select name="controlName" [formControl]="inputControl"></own-select>
```

Use `MockBuilder(Parent, ParentModule)`, render, and await stability. In separate tests, call `ngMocks.change('[name="inputName"]', 'updated')` and `ngMocks.change('[name="controlName"]', 'updated')`.

**Actual:** the targeted parent value remains `'initial'`; no error explains the disconnected operation. **Expected:** the existing mocked-binding fallback updates only the targeted parent value. This does not restore a mocked directive's complete Angular lifecycle.

## Acceptance criteria

- Track callback registration per proxy/instance and per operation; CVA metadata and the existence of `__simulateChange` do not prove registration. Inspect the [proxy](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/libs/ng-mocks/src/lib/common/mock-control-value-accessor-proxy.ts) and [inherited simulation methods](https://github.com/help-me-mom/ng-mocks/blob/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/libs/ng-mocks/src/lib/common/mock-control-value-accessor.ts).
- A registered callback remains the preferred path and equals the callback actually passed to that instance's registration method. Preserve connected real/kept/mocked controls, initial writes, touches, payload identity, and sibling isolation.
- An unregistered proxy permits the existing classic binding fallback. A custom host with no supported connection still reports a clear required/unsupported-accessor error; do not convert a null accessor into unconditional success.
- Preserve class/UI-alias identity and absence of the accessor token. Add focused source tests plus own-component spread regressions at the applicable compatibility boundaries.
- Cover the existing FormControl touch fallback as well; do not assume mocked NgModel exposes an equivalent touch contract. Track change and touch registration separately.
- Keep classic fixture APIs compatible with Angular 5–22 and use the repository's View Engine/Ivy and zoned/zoneless profiles. Put newer APIs in separate files with actual version/feature gates. Use static imports, inline setup/actions/assertions, and simple explaining comments; no helpers or runtime skips.
- Follow the existing [mocked binding examples](https://github.com/help-me-mom/ng-mocks/tree/f63b4033f7ef4195550ebfcc537df0be1c4e8a6c/examples/MockFormBindings). Use repository Docker wrappers, source coverage and affected spread validation, and keep the implementation PR CI green. Correct API/example text if the established fallback contract needs clarification.

A mocked signal `FormField` bridge is a separate issue because no equivalent existing fallback exists. Validator changes are outside scope. This issue requests follow-up implementation; no source fix is included here.

- #15033: Mocked FormField binding support — builds on this fix; signal binding semantics stay in that issue.

satanTime:
## Why

When a custom control and its classic forms directive are both mocked, `NgControl.valueAccessor` can contain a proxy attached to the correct rendered control without a registered callback. `ngMocks.change` and `ngMocks.touch` treated that proxy as connected and invoked inherited no-op methods, preventing the existing mocked-binding fallbacks from running.

## What

Track actual change and touch registrations separately on each proxy and attached instance. Prefer a registered operation, otherwise allow the local supplied `FormControl` or mocked `NgModel` change fallback. Preserve Angular's selected accessor and duplicate-accessor errors, registered callback identity and user-installed callback wrappers.

Add repository-owned regression declarations for real, kept and mocked controls, parent writes, payload and injector identity, sibling isolation, independent callback registration and unsupported custom hosts. Focused source coverage also preserves native DOM-only changes and touch events without invoking unregistered callbacks.

## Impact

Fully mocked classic bindings update the intended parent through their existing fallback instead of silently doing nothing. Connected forms and documented native DOM behavior remain unchanged. The fix does not restore a mocked directive's full forms lifecycle or add a signal-field bridge. No Material dependency or documentation change is required.

Fixes #15032.
Related to #14909. The mocked signal-field bridge remains tracked in #15033.


satanTime:

Issue #15034, opened 2026-09-14
Investigate silent mocked MatSelect changes with a null FormField accessor


satanTime:
## Problem

The [reporter’s comment on #14909](https://github.com/help-me-mom/ng-mocks/issues/14909#issuecomment-5663196874) describes mocked MatSelect leaving its parent value unchanged and pristine after `ngMocks.change('mat-select', value)`. It occurs in two components, including one without structural directives, and blocks migration of 37 existing MatSelect calls.

The reported host diagnostics are:

- A mock `__simulateChange` method exists.
- The debug element has no listeners.
- Local `NgControl` exists, but `valueAccessor` is null.
- Local `NG_VALUE_ACCESSOR` is absent.

Writing through `FormField.field()().controlValue.set(value)` updates the destination. This does **not** establish whether FormField itself was kept: a mocked directive can retain the real FieldTree as an input. The builder configuration remains unknown.

## Evidence and unresolved difference

With Angular/forms/Material 22.1.6 and ng-mocks `f63b4033`, ordinary MatSelect with real FormField works in plain TestBed, with Material kept through MockBuilder, and with Material mocked. Constructor registration, class/UI-token identity, parent writes, change, dirty state, separate touch, and sibling isolation pass.

The reporter identifies revision `b065a8b8`. Comparing it with `f63b4033` shows identical accessor lookup, mock constructor/proxy attachment, and CVA detection. The intervening #14991 native-host changes do not explain this MatSelect connection. Both revisions should throw the required-accessor error for a custom host with no accessor and no input/change listeners. The reported silent outcome therefore still needs an exact reproduction.

A different reproduced failure has a **non-null, attached but unregistered proxy** when forms directives are mocked. Track that separately; it does not reproduce the diagnostics above.

Follow-up to #14909 and its [Material comment](https://github.com/help-me-mom/ng-mocks/issues/14909#issuecomment-5663196874). These follow-ups do not assume that the commenter's exact null-accessor failure has been reproduced.

- #15030: Real CVA callback-name support.
- #15031: Mocked accessor categories.
- #15032: Disconnected mock CVA handling.
- #15033: Mocked FormField binding support.

## Investigation and acceptance

[...]

PR #15038, opened 2026-09-15
feat(forms): support mocked signal form bindings #15033

### LINKED ISSUE #15033: feat(forms): support change and touch through mocked FormField bindings

satanTime:
## Why

A mocked `FormField` preserves its input but does not connect the supplied real field to its host. Changing a mocked custom control can invoke an unregistered accessor, while a native input either fails accessor lookup or changes only its DOM value. Neither applies the requested edit to the parent field.

This feature builds on the registered-callback handling added in #15037.

## What

Add a binding-level path for `ngMocks.change` and `ngMocks.touch` when the selected host has an intentionally mocked `FormField`. Apply edits through the supplied field's `controlValue.set` and touch through `markAsTouched`, preserving Angular's debounce policy and the identities of the field tree and payload.

Resolve only an actual local binding, including a forwarded host-directive alias, without constructing unrelated providers or borrowing an ancestor's field. Preserve registered accessors, classic form fallbacks, explicit real-accessor methods and real `FormField` connections.

Add repository-owned custom CVA, model-control and native-control regressions, along with focused binding-discovery coverage. Update the existing mock-binding example, selector example, guide and change/touch API documentation. No Material dependency is added.

## Impact

Tests can use find → read → change → assert with a mocked signal-form binding and explicitly touch the field when needed. The binding path does not recreate DOM events, native value synchronization, child `writeValue` calls or model-input/output connections. Keep `FormField` real to test those Angular interactions.

Closes #15033.
Related to #14909 and #14925. The original null-accessor report remains separately tracked in #15034.


satanTime:

PR #15040, opened 2026-09-15
build(docker): clean empty browser caches after project setup

satanTime:
## Why

An interrupted Puppeteer installation can leave an empty browser build directory in the persistent cache. Later installations find the directory but fail because its executable is missing, as observed with `sh compose.sh root` and Chrome headless shell.

## What

Add one depth-limited `find` cleanup per project that uses the shared Docker browser cache. It runs immediately after the existing `docker compose up --build` step and before the explicit Puppeteer installer.

Update internal agent guidance to serialize installation runs across worktrees, including the lockfile-refresh workflow, because cleanup scans cached builds.

## Impact

Puppeteer can download a build again when a previous attempt left its directory empty. Cleanup is confined to the shared Docker volume; host and bind-mounted package-local cache directories are unchanged. Cache roots, populated builds, and browser subdirectories are preserved. Incomplete installations that already contain files still need targeted manual recovery. This changes developer installation tooling and does not trigger a package release.


satanTime:

PR #15041, opened 2026-09-15
fix(forms): support classic model control bindings #15034

### LINKED ISSUE #15034: Investigate silent mocked MatSelect changes with a null FormField accessor

satanTime:
## Why

A custom control with a classic `@Input() value` and `@Output() valueChange` can be connected to a real `FormField`, but `ngMocks.change` throws because its fallback requires signal input metadata. The same gap affects `checked`/`checkedChange`. The updated report in #15034 shows that an attached but unregistered mock accessor can expose this missing output-based route after #15032 correctly rejects its unused callbacks.

## What

- Recognize classic and signal input/output pairs through their public aliases on the same declaration, using the existing output dispatcher.
- Follow Angular's preference for `value` when a declaration also exposes `checked`.
- Add repository-owned regressions for real, kept and mocked controls, inherited aliases, object identity, field state, sibling isolation, and the reported unregistered-proxy state. Preserve registered-CVA routing and cover the Angular 21/22 touch outputs.
- Record the classic-pair contract in the internal forms investigation reference.

## Impact

`ngMocks.change` reaches the output connection that Angular already established. Registered CVAs and mocked `FormField` bindings retain their existing routes. No new API, migration or public documentation change is needed.

Fixes #15034.
Related to #14909.


satanTime:
[v14.18.1](https://github.com/help-me-mom/ng-mocks/releases/tag/v14.18.1) has been released and contains a fix for the issue. Feel free to reopen the issue or to submit a new one if you meet any problems.

